PR #21, opened 2022-07-13
'Solution'

Korenely:


KatiukhaO (changes requested):
you should to add migrations and folder with project and other base files, example, manage.py. And than tests will detect the tables from DB, because, now tests can not saw you tables

cinema/models.py
```diff
@@ -1,10 +1,28 @@
...
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
 
+    actors = models.ManyToManyField(to=Actor, related_name="movies")
+    genres = models.ManyToManyField(to=Genre, related_name="movies")
+
     def __str__(self):
```
katevasilenko, line 27: it is a best-practice to implement str magic method for all models

cinema/views.py
```diff
@@ -1,45 +1,102 @@
...
-@api_view(["GET", "POST"])
-def movie_list(request):
-    if request.method == "GET":
-        movies = Movie.objects.all()
-        serializer = MovieSerializer(movies, many=True)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data)
```
katevasilenko, line 13: I don't know is it important but it better to add status to this Response too

katevasilenko (approved):

cinema/views.py
MasterpieceElbow, line 13: Add statuses to all `Response`s

MasterpieceElbow (changes requested):

PR #30, opened 2022-07-13
Solution

RomaTsybulia:


cinema/models.py
```diff
@@ -4,7 +4,34 @@
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.TextField(max_length=255)
```
KatiukhaO, line 7: actors and genres fields in model, should be many to many fields

KatiukhaO (approved):

MasterpieceElbow (approved):

PR #38, opened 2022-07-13
SOlution

Kiawami:


cinema/models.py
```diff
@@ -1,10 +1,35 @@
...
+
+    def __str__(self):
+        return f"{self.name}"
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    genres = models.ManyToManyField(Genre)
```
KatiukhaO, line 16: model Movie should contain field actors

KatiukhaO (approved):

cinema/models.py
KatiukhaO: and , good practice it is add related name

TPolina (approved):

TPolina (changes requested):
migrations were not added, checks failed

PR #40, opened 2022-07-13
Solution

darten360:


cinema/migrations/0002_actor_cinemahall_genre_alter_movie_options_and_more.py
```diff
@@ -0,0 +1,54 @@
+# Generated by Django 4.0.2 on 2022-07-13 15:49
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cinema', '0001_initial'),
```
KatiukhaO, line 9: i see, that, you youse black for files project, may be necessary use black for migrations too, And then all file wil be in one style

KatiukhaO (approved):

TPolina (approved):

PR #41, opened 2022-07-13
Solution

NOSSUDAK:


cinema/views.py
```diff
@@ -1,45 +1,110 @@
-from rest_framework.decorators import api_view
 from rest_framework.response import Response
-from rest_framework import status
+from rest_framework import status, generics, mixins, viewsets
 
 from django.shortcuts import get_object_or_404
```
KatiukhaO, line 4: good practice when order of imports: 1. django 2. django rest 3. another

cinema/models.py
```diff
@@ -5,6 +5,32 @@ class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField("Actor")
```
KatiukhaO, line 8: good practice when add an argument "related_name" in Many To Many Relation

KatiukhaO (approved):

cinema/models.py
NOSSUDAK: I'll fix

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
+        serializer = GenreSerializer(genre, data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
```
TPolina, line 49: what if serializer is not valid?

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
TPolina, line 56: same here

TPolina (changes requested):

MasterpieceElbow (approved):

PR #88, opened 2022-11-16
Solution

mltdll:


olia-trofymchuk (approved):
As for me everything is perfect!

cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
+    path("genres/", GenreList.as_view(), name="genre-list"),
+    path(
+        "genres/<int:pk>/",
+        GenreDetail.as_view(),
+        name="movie-detail",
```
Nattalli, line 32: `genre-detail`?

cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+    ),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema_hall-list"),
```
Nattalli, line 36: also, maybe it would be better to use `_` or `-` only? not to mix them

cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(views.APIView):
+    def get_object(self, pk: int) -> Genre:
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist as e:
+            raise Http404 from e
```
Nattalli, line 38: using `get_object_or_404` would be better

Nattalli (changes requested):

cinema/urls.py
mltdll: Yeah, sometimes unchecked find-and-replace technique doesn't work that well. :)

PR #97, opened 2022-11-16
Solution

wQuelS:


ketstap162 (approved):
Looks good

cinema/models.py
```diff
@@ -1,10 +1,36 @@
 from django.db import models
 
 
+class Actor(models.Model):
+    first_name = models.CharField(max_length=100)
+    last_name = models.CharField(max_length=100)
+
+    def __str__(self):
```
Nattalli, line 8: add type annotations everywhere

cinema/urls.py
```diff
@@ -1,10 +1,42 @@
...
+    }
+)
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 
 urlpatterns = [
+    path("", include(router.urls)),
     path("movies/", movie_list, name="movie-list"),
     path("movies/<int:pk>/", movie_detail, name="movie-detail"),
```
Nattalli, line 31: hmmm... 🤔 
are you sure?

cinema/urls.py
```diff
@@ -1,10 +1,42 @@
...
+    path("genres/", GenreList.as_view(), name="genre-list"),
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema_hall-list"),
```
Nattalli, line 36: use `_` or `-`, don't mix it together

cinema/views.py
```diff
@@ -43,3 +50,109 @@ def movie_detail(request, pk):
     if request.method == "DELETE":
         movie.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+# <---
```
Nattalli, line 55: ?

cinema/views.py
```diff
@@ -43,3 +50,109 @@ def movie_detail(request, pk):
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
Nattalli, line 78: using `get_object_or_404` would be better

Nattalli (approved):
wow, Sasha, that picture is good! especially better than

Nattalli (changes requested):
sorry, I wanted to reject it

cinema/urls.py
wQuelS: > hmmm... 🤔 are you sure?

seems like i've forgotten to delete those 2 lines

cinema/views.py
wQuelS: > using `get_object_or_404` would be better
I've used this shortcut in my previous tasks. Wanted to implement something else. It is not a mistake, I guess

Nattalli: yes, that is not a mistake, just my advice. Don't think I'm being too picky about you, you just always have good code and I can't just keep approving it. I dig into the little things so that everything is perfect

wQuelS: I really appreciate the passion you apply to code reviews :)

PR #99, opened 2022-11-16
Solution

taras-andruschenko:


cinema/serializers.py
```diff
@@ -13,12 +13,77 @@ def create(self, validated_data):
...
+            "duration",
+            instance.duration
         )
-        instance.duration = validated_data.get("duration", instance.duration)
+        instance.save()
+        return instance
+
+
+class ActorSerializer(serializers.Serializer):
+    first_name = serializers.CharField(max_length=255)
```
IvanStored, line 33: i think your forgot `id` field

cinema/serializers.py
```diff
@@ -13,12 +13,77 @@ def create(self, validated_data):
...
+    last_name = serializers.CharField(max_length=255)
```
IvanStored, line 34: And actor can not exists without first and last name(same for cinema hall)

IvanStored (changes requested):

cinema/serializers.py
taras-andruschenko: I suppose "required=True" is unnecessary statement cause of i didn't define blank and null values. Same for cinema hall

IvanStored (approved):
nice!

MasterpieceElbow (approved):
GJ!

PR #101, opened 2022-11-16
Solution

pazyn404:


cinema/models.py
```diff
@@ -5,6 +5,23 @@ class Movie(models.Model):
...
+    actors = models.ManyToManyField(to="Actor", related_name="movies")
 
     def __str__(self):
         return self.title
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=63)
+    last_name = models.CharField(max_length=63)
+
```
wQuelS, line 18: It is a good practice to always override str method. Implement it everywhere

wQuelS (changes requested):
Left some comments. Also, remove cinema_serviсe_db_data.json from PR

cinema/models.py
pazyn404: Thanks for review, but I think cinema_data should be presented in the repository :)

wQuelS (approved):

Nattalli (approved):
👍🏻

PR #104, opened 2022-11-17
Solution

VadymShkarbul:


cinema/serializers.py
```diff
@@ -17,7 +17,64 @@ def update(self, instance, validated_data):
...
+        instance.name = validated_data.get("name", instance.name)
+        instance.save()
+
+        return instance
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=255, required=True)
+    last_name = serializers.CharField(max_length=255, required=True)
```
skrypka-dmytro, line 46: https://www.django-rest-framework.org/api-guide/fields/#required
fix everywhere

cinema/views.py
```diff
@@ -1,45 +1,115 @@
...
-        movies = Movie.objects.all()
-        serializer = MovieSerializer(movies, many=True)
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+
+class GenreList(APIView):
+    @staticmethod
+    def get(request):
```
skrypka-dmytro, line 21: Why are you using @staticmethod?

skrypka-dmytro (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,115 @@
...
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    @staticmethod
+    def get(request, pk):
```
skrypka-dmytro, line 38: DRY, make a method that will take a genre

skrypka-dmytro (changes requested):

cinema/serializers.py
VadymShkarbul: Thank you for the review! Yes you are right required=True is default option, so not needable here.

cinema/views.py
VadymShkarbul: I'm using @staticmethod because in both methods we are not accessing instance of the class

cinema/views.py
VadymShkarbul: Yes I fix it

skrypka-dmytro (approved):
GJ!

cinema/admin.py
```diff
@@ -1,8 +1,23 @@
...
+
+@admin.register(Genre)
+class GenreAdmin(admin.ModelAdmin):
+    pass
+
+
+@admin.register(Actor)
+class ActorAdmin(admin.ModelAdmin):
+    pass
+
```
Nattalli, line 19: you can use `admin.site.register(ModelName)` and it would do the same, but without `pass` station

cinema/models.py
```diff
@@ -1,10 +1,36 @@
 from django.db import models
 
 
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __str__(self):
```
Nattalli, line 8: where is a type annotation? 🤔

Nattalli: add it everywhere

cinema/views.py
```diff
@@ -1,45 +1,118 @@
...
+    def get_object(pk):
+        return get_object_or_404(Genre, pk=pk)
 
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+    def get(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre)
+        return Response(serializer.data)
```
Nattalli, line 44: add status

cinema/views.py
```diff
@@ -1,45 +1,118 @@
...
+    def delete(self, request, pk):
+        genre = self.get_object(pk)
+        genre.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(mixins.ListModelMixin,
+                mixins.CreateModelMixin,
+                generics.GenericAPIView
+                ):
```
Nattalli, line 78: fix code style. Remember:
Good:
```python
class A(
...
)
```
Also good:
```python
class A(...)
```
Bad:
```python
class A(...
)
```

Nattalli (changes requested):

cinema/views.py
VadymShkarbul: Thanks for the review! Concentrated on one thing - lost another. All fixed ;)

Nattalli (approved):
Nice!

PR #107, opened 2022-11-17
Solution

Vasyl-Poremchuk:


cinema/serializers.py
```diff
@@ -1,6 +1,63 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=255, required=True)
```
linaSam, line 8: "required=True" is default, no need to write it

linaSam (changes requested):

PR #108, opened 2022-11-17
Develop

alexzubkoff:


cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
         return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self):
+        return f"{self.name} {self.rows} {self.seats_in_row}"
```
mltdll, line 36: If you are going to show those numbers in the string representation of the object, consider adding some kind of indication for what they represent. For example:
```
Cinema hall 1 (rows: 10, seats in row:5)
```
But it might be better not to display them here at all.

cinema/urls.py
```diff
@@ -1,10 +1,45 @@
...
 
-from cinema.views import movie_list, movie_detail
+from cinema.views import (
+    movie_list,
+    movie_detail,
+    GenreList,
+    GenreDetail,
+    ActorList,
+    ActorDetail,
+    CinemaHallViewSet, MovieViewSet,
```
mltdll, line 11: Move the second item to the next line.

cinema/views.py
```diff
@@ -40,6 +48,91 @@ def movie_detail(request, pk):
...
+
     if request.method == "DELETE":
         movie.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class GenreList(APIView):
+
+    def get(self, request):
+        genres = Genre.objects.all()[:20]
```
mltdll, line 67: Slicing is not the best idea here. The problem of getting too much data can be solved using pagination, which we will soon encounter. As for now, I'd suggest leaving it as is, i.e. `Genre.objects.all()`.

cinema/views.py
```diff
@@ -40,6 +48,91 @@ def movie_detail(request, pk):
...
+
+
+class ActorList(generics.ListCreateAPIView):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class ActorDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
```
mltdll, line 121: > * For the Actor model use a GenericAPIView

Your solution is not worse but is not quite what was asked.

mltdll (changes requested):

cinema/views.py
alexzubkoff: Thanks for good advices!

cinema/models.py
```diff
@@ -33,4 +33,6 @@ class CinemaHall(models.Model):
     seats_in_row = models.IntegerField()
 
     def __str__(self):
-        return f"{self.name} {self.rows} {self.seats_in_row}"
+        return f"Cinema hall: {self.name} " \
+               f"(rows: {self.rows}, " \
+               f"seats in row: {self.seats_in_row})"
```
mltdll, line 38: Using `()` is the preferred style for line breaking.

mltdll (approved):

cinema/models.py
```diff
@@ -1,10 +1,38 @@
 from django.db import models
 
 
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __str__(self):
```
Nattalli, line 8: what about _old but gold_ type annotation? 😺

cinema/models.py
```diff
@@ -1,10 +1,38 @@
...
+        return f"{self.first_name} {self.last_name}"
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255, unique=True, null=True)
```
Nattalli, line 13: it can't be `null`. If at least two names would be `null` - then an error would occur as objects aren't unique

cinema/views.py
```diff
@@ -40,6 +48,91 @@ def movie_detail(request, pk):
...
         movie.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class GenreList(APIView):
+
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data)
```
Nattalli, line 69: add status

Nattalli (changes requested):

cinema/models.py
Nattalli: yes, please, fix your code style

cinema/models.py
```diff
@@ -5,14 +5,14 @@ class Actor(models.Model):
...
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
 
 
 class Genre(models.Model):
-    name = models.CharField(max_length=255, unique=True, null=True)
+    name = models.CharField(max_length=255, unique=True)
```
Nattalli, line 13: I guess that you had to do migrations and push it

Nattalli (changes requested):

PR #110, opened 2022-11-17
Solution for the task

olia-trofymchuk:


cinema/serializers.py
```diff
@@ -18,7 +18,56 @@ def update(self, instance, validated_data):
             "description", instance.description
         )
         instance.duration = validated_data.get("duration", instance.duration)
+        instance.save()
+        return instance
+
+
+class ActorSerializer(serializers.Serializer):
```
mltdll, line 25: I think it would be better to use `ModelSerializer`s here. But I guess you already did it in the next task, so it's probably not necessary to fix it here as well. _I guess._

mltdll (approved):

cinema/models.py
```diff
@@ -1,10 +1,36 @@
 from django.db import models
 
 
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __str__(self):
```
Nattalli, line 8: add type annotations here and everywhere

cinema/serializers.py
Nattalli: in this task, it's not a mistake. but I prefer using ModelSerializer

cinema/serializers.py
```diff
@@ -18,7 +18,56 @@ def update(self, instance, validated_data):
...
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(required=True, max_length=255)
+    last_name = serializers.CharField(required=True, max_length=255)
+
```
Nattalli, line 29: https://www.django-rest-framework.org/api-guide/fields/#required
it's the default value, so it's unnecessary

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
Nattalli, line 35: using `get_object_or_404` would be better

Nattalli (changes requested):

Nattalli (approved):

PR #130, opened 2022-12-29
Implement endpoints through different class based views

Kobsar703:


YuriiKindrat (changes requested):
all tests should pass

YuriiKindrat (approved):

PR #161, opened 2023-03-17
Solution

YuriiVataschuk:


ArsenPidhoretskyi (changes requested):
Migrations does not exist

cinema/views.py
```diff
@@ -1,45 +1,116 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
Vasyl-Smutok, line 36: You can use `get_or_404` instead of `try-except`

Vasyl-Smutok (changes requested):

Grayder0152 (approved):
Pls, add typing to methods

PR #172, opened 2023-04-02
Solution

MaybeTI:


DmytryTong (approved):
If it passes the tests and work, then - fine)

PR #176, opened 2023-04-04
Solution

SlavikSherbak:


cinema/urls.py
```diff
@@ -1,10 +1,38 @@
...
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("genres/", GenreList.as_view(), name="genres-list"),
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-halls"),
+    path("cinema_halls/<int:pk>/", cinema_hall_detail, name="cinema-halls"),
+    path("", include(router.urls))
```
ArsenPidhoretskyi, line 35: No need to include if base path does not provided. Instead use:
urlpatterns = [list of patterns] + router.urls

cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
```
ArsenPidhoretskyi, line 61: Instead of using if you can use:
`serializer.is_valid(raise_exception=True)`
Fix in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #189, opened 2023-04-05
solution

amber-marichi:


yura-hudzovskyi (approved):
GJ!

cinema/serializers.py
```diff
@@ -1,13 +1,13 @@
...
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
 
 
 class MovieSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
-    title = serializers.CharField(max_length=255, required=True)
-    description = serializers.CharField(required=True)
-    duration = serializers.IntegerField(required=True)
+    title = serializers.CharField(max_length=255)
```
Y-Havryliv, line 8: thanks for fixing the source code)

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
-def movie_list(request):
-    if request.method == "GET":
-        movies = Movie.objects.all()
-        serializer = MovieSerializer(movies, many=True)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+class GenreList(APIView):
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data)
```
Y-Havryliv, line 20: status code

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
+    def get_object(self, pk):
+        return get_object_or_404(Genre, pk=pk)
 
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+    def get(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre)
+        return Response(serializer.data)
```
Y-Havryliv, line 37: make sure you have statuses everywhere

Y-Havryliv (changes requested):

cinema/serializers.py
amber-marichi: haha, sorry :) pretty sure i was doing it before submitting , probably havnt even realised that i had been originally that way, that it isn't something that i messed up as usual. will change it back

cinema/views.py
amber-marichi: i definitely read that Response returns status 200 by default, thought that is not necessary. Ill add them, ty.

cinema/serializers.py
Y-Havryliv: no need, source code is wrong)

cinema/views.py
Y-Havryliv: Explicit is better than implicit.

Y-Havryliv (approved):
Well done!

PR #234, opened 2023-05-09
py-api-views

OmeLchDO:


OmeLchDO:
Why can't I use the @staticmethod?

Grayder0152 (approved):

PR #235, opened 2023-05-12
done

phaishuk:


cinema/views.py
```diff
@@ -1,45 +1,106 @@
...
+        serializer = GenreSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        return get_object_or_404(Genre, pk=pk)
```
olena-hul, line 30: You can just use get_object_or_404 directly, no need to move 1 line into a separate method

olena-hul (changes requested):

PR #239, opened 2023-05-25
Solution

Amirammi:


cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+
         if serializer.is_valid():
```
vasylhnatiuk, line 25: use serializer.is_valid(raise_exception=True) instead of if-else here and oall of the code

vasylhnatiuk (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
+        if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+
+        if serializer.is_valid(raise_exception=True):
```
ArsenPidhoretskyi, line 51:
```suggestion
        serializer.is_valid(raise_exception=True):
```
If is not needed. Fix in all places

ArsenPidhoretskyi (changes requested):

PR #251, opened 2023-06-14
Solution

AlexandrZhezheria:


cinema/urls.py
```diff
@@ -1,10 +1,26 @@
...
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("genres/", GenreList.as_view(), name="genre-list"),
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("", include(router.urls)),
+    path("", include(router.urls))
```
ArsenPidhoretskyi, line 23: 1. Why you do include twice?
2. Also no need to do include if basepath is not provided instead use:
urlpatters = [list of urls] + router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #252, opened 2023-06-17
Solution_v1.0

bohdan-yatsyna:


cinema/urls.py
```diff
@@ -1,10 +1,40 @@
...
+    "patch": "partial_update",
+    "delete": "destroy",
+})
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi, line 27: No need to include if basepath is not provided. Instead do:
urlpatters = [list of urls] + router.urls

cinema/views.py
```diff
@@ -1,45 +1,124 @@
...
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
-        if serializer.is_valid():
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
```
ArsenPidhoretskyi, line 47: if is redundant. Check how  raise_exception=True works. Fix in all places
```suggestion
        serializer.is_valid(raise_exception=True):
```

ArsenPidhoretskyi (changes requested):

cinema/views.py
bohdan-yatsyna: Thank you for the remarks!

ArsenPidhoretskyi (approved):
Well done!

PR #255, opened 2023-06-22
Solution

Roman28101:


cinema/views.py
```diff
@@ -1,45 +1,123 @@
...
-        serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
vasylhnatiuk, line 34: you don't need it if you use if serializer.is_valid(raise_exception=True):

vasylhnatiuk: chenge it in all of the code

vasylhnatiuk (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk, line 30: you don't need `if `because if form will be not valid, function will return error

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #263, opened 2023-06-29
Solution

andriy-md:


cinema/urls.py
```diff
@@ -1,10 +1,38 @@
...
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("genres/", GenreList.as_view(), name="genre-list"),
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema-halls/", cinema_hall_list, name="cinema-hall-list"),
+    path("cinema-halls/<int:pk>/", cinema_hall_detail, name="cinema-hall-detail"),
+    path("", include(router.urls))
```
ArsenPidhoretskyi, line 35: No need to include if basepath is not provided. Instead use:

cinema/views.py
```diff
@@ -1,45 +1,116 @@
...
+class CinemaHallViewSet(viewsets.ViewSet):
+    def list(self, request):
+        queryset = CinemaHall.objects.all()
+        serializer = CinemaHallSerializer(queryset, many=True)
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def create(self, request):
+        serializer = CinemaHallSerializer(data=request.data)
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 81: No need to use if. Instead use:

ArsenPidhoretskyi: Fix in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #264, opened 2023-06-29
Solution

Sebshe:


cinema/urls.py
```diff
@@ -1,10 +1,45 @@
...
+    }
+)
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi, line 32: No need to include if basepath is not provided instead use:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #265, opened 2023-07-02
Solution

xtina999:


cinema/urls.py
```diff
@@ -1,10 +1,24 @@
...
 
-from cinema.views import movie_list, movie_detail
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet, basename="cinema_hall")
+router.register("movies", MovieViewSet, basename="movie")
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
vsmutok, line 17: No need to include if basepath is not provided. Instead use:

cinema/views.py
```diff
@@ -1,45 +1,119 @@
...
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 112: No need to use if. Instead use:

vsmutok: Fix in all places

cinema/views.py
```diff
@@ -1,45 +1,119 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
vsmutok, line 91: use `get_object_or_404`

vsmutok (changes requested):

vsmutok (approved):
well done 👍

PR #276, opened 2023-07-14
Solution

olenaliuby:


cinema/models.py
```diff
@@ -1,10 +1,40 @@
...
         return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int | float:
```
ArsenPidhoretskyi, line 36: How it can return float?

cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy",
+})
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", APIRoot.as_view(), name="api-root"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi, line 30: No need to include if basepath is not provided. Instead do:

cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+    "get": "retrieve",
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy",
+})
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", APIRoot.as_view(), name="api-root"),
```
ArsenPidhoretskyi, line 29: Why you need this??

ArsenPidhoretskyi (changes requested):

cinema/urls.py
olenaliuby: I created APIRoot view in order to have all links included at Api Root: 
Screenshot 2023-07-15 at 13 51 16

Is it a bad practice?

vasylhnatiuk: this task does not need it, but it is not mistake

vasylhnatiuk (approved):

PR #284, opened 2023-07-26
done

nazarbychkar:


cinema/views.py
```diff
@@ -1,45 +1,91 @@
...
             serializer.save()
-            return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(serializer.data)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: Use serializer.is_valid(raise_exception=True) which will automatically return 400 if the serializer is not valid

vsmutok: Fix everywhere

cinema/urls.py
```diff
@@ -1,10 +1,45 @@
...
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
```
vsmutok: No need to include if basepath is not provided. Instead use:

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #305, opened 2023-07-30
Solution

Andriy-Sydorenko:


cinema/urls.py
```diff
@@ -1,10 +1,29 @@
...
+
+from cinema.views import (MovieViewSet,
+
+                          GenreList,
+                          GenreDetail,
+
+                          ActorList,
+                          ActorDetail,
+
+                          CinemaHallViewSet)
```
YuriiKindrat: why do u need blank lines here?

YuriiKindrat (changes requested):

cinema/urls.py
Andriy-Sydorenko: to separate groups of imports, so the different views are grouped by model they belong to

Oleg-s-s (approved):

PR #310, opened 2023-08-01
implemented class views

mariiahorbova:


YuriiKindrat (changes requested):
all tests should pass

YuriiKindrat (approved):

PR #319, opened 2023-08-11
Solution

Niki-Alex:


cinema/serializers.py
```diff
@@ -1,16 +1,52 @@
...
+    )
 
     def create(self, validated_data):
-        return Movie.objects.create(**validated_data)
+        actors_data = validated_data.pop("actors", [])
+        genres_data = validated_data.pop("genres", [])
+        movie = Movie.objects.create(**validated_data)
+
+        for actor in actors_data:
+            movie.actors.add(actor)
```
ArsenPidhoretskyi: Use .set

cinema/urls.py
```diff
@@ -1,10 +1,40 @@
...
+    "delete": "destroy"
+})
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include if basepath is not provided. Instead use:

cinema/views.py
```diff
@@ -1,45 +1,134 @@
...
+            serializer.save()
+            return Response(
+                serializer.data,
+                status=status.HTTP_200_OK
+            )
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
ArsenPidhoretskyi: No need to use if-statement. Instead use serializer.is_valid(raise_exception=True)

pytest.ini
ArsenPidhoretskyi, line 1: Restore this file

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #321, opened 2023-08-14
Solution

Oleksiy-Lyashenko:


cinema/models.py
```diff
@@ -1,10 +1,39 @@
...
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField(
+        Actor,
+        related_name="movie"
```
pavlejviki: the related name should be in plural better

cinema/serializers.py
```diff
@@ -1,6 +1,72 @@
...
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Actor
+        fields = "__all__"
+
+    def create(self, validated_data):
+        return Actor.objects.create(**validated_data)
+
```
pavlejviki: it's not essential to define create and update methods  if  you don't need any custom behavior. The default create and update methods provided by serializers.ModelSerializer would essentially look like what you've written.

cinema/serializers.py
```diff
@@ -1,6 +1,72 @@
...
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
+
+    def create(self, validated_data):
+        return Genre.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
```
pavlejviki: view comment above

cinema/serializers.py
```diff
@@ -1,6 +1,72 @@
...
+            instance.name)
+        instance.rows = validated_data.get(
+            "rows",
+            instance.rows)
+        instance.seats_in_row = validated_data.get(
+            "seats_in_row",
+            instance.seats_in_row)
+        instance.save()
+
+        return instance
```
pavlejviki: view comment above

cinema/views.py
```diff
@@ -1,45 +1,102 @@
...
+    def get_genre_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
 
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
+    def get(self, request, pk):
+        genre = self.get_genre_object(pk)
```
pavlejviki, line 40: you  can use the  get_object_or_404 method from django.shortcuts

cinema/views.py
```diff
@@ -1,45 +1,102 @@
...
+        serializer = GenreSerializer(genre)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk):
+        genre = self.get_genre_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
         if serializer.is_valid():
```
pavlejviki: u can simplify your code  using raise_exception=True argument with serializer.is_valid(). When you set raise_exception=True, if the serializer is not valid, it will automatically raise a serializers.ValidationError that DRF will handle by returning a 400 Bad Request response with the serializer's errors.

But it's up to your preferences

pavlejviki (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,94 @@
...
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_genre_object(self, pk):
+        return get_object_or_404(Genre, pk=pk)
```
pavlejviki, line 37: For developers familiar with Django, get_object_or_404 is a recognizable function. Wrapping it might make someone wonder if there's custom logic involved

cinema/models.py
```diff
@@ -1,10 +1,39 @@
...
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField(
+        Actor,
+        related_name="movies"
+    )
+    genres = models.ManyToManyField(
+        Genre,
+        related_name="genres"
```
pavlejviki: if you have a Genre object, you'd use genre_object.genres.all() to get all the movies of that genre. This can be confusing, as the naming suggests you're getting genres, not movies.

so better to use related_name="movies", the same as for actors

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #322, opened 2023-08-17
Solution

Sashkoooo:


cinema/urls.py
```diff
@@ -1,10 +1,39 @@
...
+    "get": "retrieve",
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy",
+})
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include when basepath is not provided. Instead use:

ArsenPidhoretskyi (changes requested):

cinema/migrations/0002_actor.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.1 on 2023-08-15 19:09
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cinema', '0001_initial'),
```
vasylhnatiuk: change ' to ", you can use Black

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #336, opened 2023-09-25
Solution

VladyslavKlevanskyi:


cinema/views.py
```diff
@@ -1,45 +1,118 @@
...
-        serializer = MovieSerializer(movie)
-        return Response(serializer.data, status=status.HTTP_200_OK)
-
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
-        if serializer.is_valid():
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
EdAlekseiev: You don't need to use if-statement if you use `is_valid` method with `raise_exception=True`

cinema/views.py
```diff
@@ -1,45 +1,118 @@
...
             serializer.save()
-            return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(serializer.data)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
```
EdAlekseiev: The same comment as above

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #340, opened 2023-10-05
'Solution'

Mazur-Mariia-Python:


cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
-
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
-        if serializer.is_valid():
-            serializer.save()
-            return Response(serializer.data, status=status.HTTP_200_OK)
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        serializer.is_valid()
```
EdAlekseiev: Pass argument raise_exception=True to is_valid method.

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+        serializer.save()
+        return Response(serializer.data)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        serializer.is_valid()
```
EdAlekseiev: The same comment as above

EdAlekseiev (changes requested):

cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy",
+})
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include if basepath is not provided. Instead use:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #341, opened 2023-10-05
solution

VladyslavOladko:


cinema/models.py
```diff
@@ -1,10 +1,46 @@
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(
+        Actor,
+        related_name="actors"
```
ArsenPidhoretskyi: wrong related name

cinema/models.py
```diff
@@ -1,10 +1,46 @@
...
+    )
+    genres = models.ManyToManyField(
+        Genre,
+        related_name="genres"
```
ArsenPidhoretskyi: wrong related name

cinema/urls.py
```diff
@@ -1,10 +1,45 @@
...
+        "cinema_halls/",
+        cinema_hall_list,
+        name="cinema-hall-list"
+    ),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include if basename is not provided. Instead use:

cinema/views.py
```diff
@@ -1,45 +1,119 @@
...
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk):
+        serializer = GenreSerializer(
+            self.get_object(pk=pk),
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Instead of if-statement use:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #344, opened 2023-10-09
implemented solution

innyshka:


cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
+    serializer_class = ActorSerializer
+
+    def get(self, request, *args, **kwargs):
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs):
+        return self.update(request, *args, partial=True, **kwargs)
```
eduardhabryd: here should be used `self.partial_update(request, *args, **kwargs)` method

eduardhabryd (changes requested):
Good in general but you should use partial_update with patch request

nyanplague (approved):

Serhii-Leonenko (approved):

PR #357, opened 2023-10-11
Implemented API

ORENYT:
Created api for all models with different serializers


cinema/models.py
```diff
@@ -5,6 +5,23 @@ class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField("Actor")
+    genres = models.ManyToManyField("Genre")
```
Ihor-MA, line 9: You can add related name

cinema/serializers.py
```diff
@@ -1,24 +1,27 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
```
Ihor-MA, line 3: You can also use relative imports, IMHO it is better

Ihor-MA (approved):

cinema/serializers.py
ORENYT: relative imports are a very bad practice)

cinema/models.py
ORENYT: for what reason? They are not used

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Serhii-Leonenko, line 52: You can simplify it with raise_exceptions = True https://stackoverflow.com/a/40373171

Serhii-Leonenko (approved):

PR #363, opened 2023-10-12
Solution

Kurta97:


cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Sofbog: follow the same code style,  `serializer.is_valid(raise_exception=True)`

Sofbog (changes requested):

maxkatkalov (approved):

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        genre = Genre.objects.all()
+        serializer = GenreSerializer(genre, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
Sofbog: you have this place 
```
    def put(self, request, pk):
        genre = self.get_object(pk)
        serializer = GenreSerializer(genre, data=request.data)
        serializer.save()
```
why do you ned if block here?

Sofbog (changes requested):

cinema/views.py
Serhii-Leonenko: fix this according to the Bogdan's comment, the same in other similar cases

Serhii-Leonenko (changes requested):

cinema/views.py
Kurta97: I have fixed it

Serhii-Leonenko: no, you haven't. You don't need if condition when you are using raise_exception=True. 

image

Serhii-Leonenko (changes requested):

Sofbog (approved):
LGTM

PR #365, opened 2023-10-12
Solution

OstapIvakh:


cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Sofbog: in all places like this you can refactor to DRF documentation using

Sofbog (changes requested):

cinema/views.py
```diff
@@ -22,7 +22,7 @@ def get(self, request):
     def post(self, request):
         serializer = GenreSerializer(data=request.data)
 
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Serhii-Leonenko: this line is redundant if you use raise_exception=True, the same in all other similar cases

Serhii-Leonenko (changes requested):

cinema/views.py
```diff
@@ -25,7 +25,6 @@ def post(self, request):
         if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
```
Serhii-Leonenko: You don't need` if `condition at all if you are using raise_exception=True

Serhii-Leonenko (changes requested):

Sofbog (approved):
LGTM

PR #369, opened 2023-10-12
Solution

anna-lybid:


cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Sofbog: in all places like this, you can refactor follow-up the DRF documentation using

Sofbog (changes requested):

Serhii-Leonenko (approved):

PR #370, opened 2023-10-13
Done

maxkatkalov:


cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
+    def post(self, request: Request) -> Response:
+        serialized_data = GenreSerializer(data=request.data)
+        if serialized_data.is_valid():
+            serialized_data.save()
+            return Response(
+                serialized_data.data, status=status.HTTP_201_CREATED
+            )
+        return Response(
+            serialized_data.errors, status=status.HTTP_400_BAD_REQUEST
+        )
```
Sofbog: in all places like this, you can refactor following the DRF documentation using

Sofbog (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
+    def post(self, request: Request) -> Response:
+        serialized_data = GenreSerializer(data=request.data)
+        if serialized_data.is_valid(raise_exception=True):
+            serialized_data.save()
+            return Response(
+                serialized_data.data, status=status.HTTP_201_CREATED
+            )
+        return Response(
+            serialized_data.errors, status=status.HTTP_400_BAD_REQUEST
+        )
```
Sofbog: you do not need if block and raise exception when you are using serialized_data.is_valid(raise_exception=True)

Sofbog (changes requested):

Sofbog (approved):
LGTM

PR #374, opened 2023-10-16
API Views

IProskurnytskyi:


cinema/views.py
```diff
@@ -1,45 +1,116 @@
...
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(
+            instance=genre,
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
viktoria-rybenchuk: simplify your code with raise_exception=True https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true

viktoria-rybenchuk (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk: int) -> Genre:
+        return get_object_or_404(Genre, id=pk)
```
Dimosphen1, line 32: You could use `get_object_or_404` directly in your methods where needed, there is no need to allocate a separate method for that.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #376, opened 2023-10-25
Solution

nickkozlov90:


cinema/urls.py
```diff
@@ -1,10 +1,40 @@
...
+    "delete": "destroy"
+})
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router if basepath is not provided. Instead use:

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Good job!

PR #377, opened 2023-10-30
Develop

vlodek69:


cinema/urls.py
```diff
@@ -1,10 +1,41 @@
...
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include a router when basepath is not provided. Instead use:

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
         if serializer.is_valid():
             serializer.save()
-            return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(serializer.data)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, request.data, partial=True)
+        if serializer.is_valid():
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #387, opened 2023-11-19
implemented py-api-views task

HlovaOleh:


cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+    "patch": "partial_update",
+    "delete": "destroy",
+})
 
-from cinema.views import movie_list, movie_detail
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router when baseurl is not provided. Use: urlpatterns = [list of urls] + router.urls

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #393, opened 2023-11-28
Solution

prochigor:


cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+
+    def __str__(self) -> str:
+        return str(self.name)
```
petrykivd, line 16: Do you really need here str()?

cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name} rows: {self.rows} seats: {self.seats_in_row}"
```
petrykivd, line 25: Perhaps it is more logical to return the name of the hall?

cinema/serializers.py
```diff
@@ -22,3 +28,24 @@ def update(self, instance, validated_data):
...
+    class Meta:
+        model = Actor
+        fields = ["id", "first_name", "last_name"]
+
+
+class CinemaHallSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = CinemaHall
+        fields = ["id", "name", "rows", "seats_in_row"]
```
petrykivd: You can use fields = "__all__"

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
-
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
+class GenreDetail(APIView):
+    def get(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
```
petrykivd: You can create get_object method, It will be more readable.

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        serializer = GenreSerializer(genre)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
```
petrykivd: Same here

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        serializer = GenreSerializer(genre, data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = get_object_or_404(Genre, pk=pk)
```
petrykivd: and next

petrykivd (changes requested):

cinema/views.py
prochigor: I'm not sure that it makes a sence creating a method for 1 line, becouse it only increase amount of code, but ok

cinema/serializers.py
prochigor: it is a good point

cinema/models.py
prochigor: it depends on your task, but it is not important here, and to my mind it is better like that

cinema/models.py
prochigor: This is because I add an annotation type, and it has an underline without it

cinema/models.py
prochigor: I don't agree, it depends on the task you have, but here is not business logic about that.

Y-Havryliv (changes requested):
tests should pass

prochigor:
I have just pushed that(

cinema/serializers.py
```diff
@@ -1,24 +1,59 @@
...
+        return movie
 
     def update(self, instance, validated_data):
         instance.title = validated_data.get("title", instance.title)
         instance.description = validated_data.get(
             "description", instance.description
         )
         instance.duration = validated_data.get("duration", instance.duration)
-
+        actors_data = validated_data.pop("actors", "")
```
Y-Havryliv: it's better to use an empty list instead of string

Y-Havryliv (approved):

PR #399, opened 2023-11-29
Implement py-api-views

SevKrok:


RabenkoYevhenii (approved):
Looks correct)

cinema/serializers.py
LiudmylaKulinchenko, line 1: Why don't u use ModelSerializer instead?
You can also skip rewriting previous logic

LiudmylaKulinchenko (changes requested):

Y-Havryliv (approved):
gj!

PR #403, opened 2023-11-29
Develop

Familenko:


cinema/serializers.py
```diff
@@ -1,24 +1,113 @@
...
+
+        movie = Movie.objects.create(**validated_data)
+        movie.actors.set(actors)
+        movie.genres.set(genres)
+
+        return movie
 
     def update(self, instance, validated_data):
+        actors_data = validated_data.pop("actors", None)
+        genres_data = validated_data.pop("genres", None)
```
prochigor: it's better to use an empty list instead of none

cinema/serializers.py
```diff
@@ -1,24 +1,113 @@
...
+
         instance.title = validated_data.get("title", instance.title)
         instance.description = validated_data.get(
             "description", instance.description
         )
         instance.duration = validated_data.get("duration", instance.duration)
 
+        is_partial = self.context["request"].method == "PATCH"
```
prochigor: do you need this here? update method work only with methods put, patch

cinema/serializers.py
```diff
@@ -1,24 +1,113 @@
...
+
+        if actors_data:
+            if is_partial:
+                instance.actors.add(*actors_data)
```
prochigor: you can use only set here, without conditions

prochigor (changes requested):

cinema/serializers.py
Familenko: Yes, update method work only with methods put, patch. And depends on that wich method in context this code will renew all information with method set or put additional information with method add

cinema/serializers.py
Familenko: no

cinema/serializers.py
Familenko: Why?

prochigor: mentors write it to me)

Familenko: good argument)

cinema/serializers.py
```diff
@@ -1,24 +1,113 @@
...
     id = serializers.IntegerField(read_only=True)
     title = serializers.CharField(max_length=255)
     description = serializers.CharField()
     duration = serializers.IntegerField()
+    actors = serializers.PrimaryKeyRelatedField(
+        queryset=Actor.objects.all(), many=True
+    )
+    genres = serializers.PrimaryKeyRelatedField(
+        queryset=Genre.objects.all(), many=True
+    )
```
LiudmylaKulinchenko: you were not supposed to do so in **_this_** task according to the description
image

cinema/serializers.py
```diff
@@ -1,24 +1,113 @@
...
+            "first_name",
+            instance.first_name
+        )
+        instance.last_name = validated_data.get(
+            "last_name", instance.last_name
+        )
+
+        instance.save()
+
+        return instance
```
LiudmylaKulinchenko, line 46: this code ( as well as similar code in the next serializers) will not be needed if you do everything according to the task description. Fix everywhere

LiudmylaKulinchenko (changes requested):

cinema/serializers.py
Familenko: It was interesting to make this investigation. Ok, I will delete all this code

Y-Havryliv (approved):

PR #404, opened 2023-11-29
Solution

YuliiaSaliuta:


Shatrovskyi (approved):
Well done!

cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+        "cinema_halls/",
+        cinema_hall_list,
+        name="cinemahall-list",
+    ),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinemahall-detail",
+    ),
+    path("", include(router.urls))
```
wQuelS: it is overall better to put such general path at the beginning

cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path(
+        "cinema_halls/",
+        cinema_hall_list,
+        name="cinemahall-list",
+    ),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinemahall-detail",
```
wQuelS: Yeah, you were trying to avoid `-`. But there is no such word as `cinemahall`. Try to come up with solution that will use `cinema hall` and also will not mix `_` and `-`.

wQuelS (changes requested):

cinema/urls.py
```diff
@@ -24,21 +24,21 @@
...
+    path("", include(router.urls)),
     path("genres/", GenreList.as_view(), name="genre-list"),
     path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
     path("actors/", ActorList.as_view(), name="actor-list"),
     path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
     path(
         "cinema_halls/",
         cinema_hall_list,
-        name="cinemahall-list",
+        name="hall-list",
```
wQuelS, line 35: nice :joy::joy:

I meant something like `cinema-hall`. But let it be))

wQuelS (approved):
The skill to avoid unnecessary tough decisions is pretty valuable. Nicely done

PR #407, opened 2023-11-29
Develop

RVChornyy:


Naz-iv (approved):

cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+    name = models.CharField(max_length=63, unique=True)
+
+    def __str__(self):
+        return self.name
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=63)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
```
NEkropo1: I think we should check if someone passes a negative value, just for stability in our db

cinema/serializers.py
```diff
@@ -22,3 +22,60 @@ def update(self, instance, validated_data):
...
+        )
+
+        instance.save()
+
+        return instance
+
+
+class GenreSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    name = serializers.CharField(max_length=63, required=True)
```
NEkropo1, line 52: Any other fields?

cinema/serializers.py
```diff
@@ -22,3 +22,60 @@ def update(self, instance, validated_data):
...
+
+        instance.save()
+
+        return instance
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=63, required=True)
+    last_name = serializers.CharField(max_length=63, required=True)
```
NEkropo1, line 68: Any other fields here or in movies?)

NEkropo1 (changes requested):

cinema/serializers.py
RVChornyy: It's been noticed in Readme, that we must not use related serializers for ManyToMany relations. For genres and actors there are ManyToMany relations and if I use extra field for Genre or Actor Serializer, I need to implement another Serializer for list or retrieve action, but we have no such a task here. Please, advice, what fields should i add there?

PR #408, opened 2023-11-29
solution

Naz-iv:


roffi37 (approved):

cinema/serializers.py
```diff
@@ -1,13 +1,27 @@
...
+        fields = "__all__"
 
 
 class MovieSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
     title = serializers.CharField(max_length=255)
     description = serializers.CharField()
     duration = serializers.IntegerField()
+    actors = ActorSerializer(many=True, read_only=True)
+    genres = GenreSerializer(many=True, read_only=True)
```
NEkropo1: Read the task, please. You shouldn't use related serializers.

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        return get_object_or_404(Genre, pk=pk)
 
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
+    def get(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre)
         return Response(serializer.data, status=status.HTTP_200_OK)
```
NEkropo1, line 43: try get_object_or_404

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def delete(self, request, pk):
+        genre = self.get_object(pk)
```
NEkropo1, line 62: what if we wont find requested object?

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def put(self, request, pk):
+        genre = self.get_object(pk)
```
NEkropo1, line 54: what if we wont find requested object?

NEkropo1 (changes requested):

cinema/views.py
Naz-iv: this is why I have this method in lines 37-38...     
def get_object(self, pk):
        return get_object_or_404(Genre, pk=pk)

cinema/views.py
Naz-iv: this is why I have this method in lines 37-38...
def get_object(self, pk):
return get_object_or_404(Genre, pk=pk)

cinema/views.py
Naz-iv: this is why I have this method in lines 37-38...
def get_object(self, pk):
return get_object_or_404(Genre, pk=pk)

Y-Havryliv (approved):

cinema/views.py
NEkropo1: Good idea, but nested calling without reason only makes code more complex. Better not to do that if it is not clearly defined in business logic. Don't you think there is some idea behind namespace `get_object_or_404` ?)
I mean I missed it, my mistake, but still, better not to do like this)

PR #410, opened 2023-11-29
solution

roffi37:


DenPrislipskyi (approved):

cinema/models.py
```diff
@@ -1,9 +1,32 @@
 from django.db import models
 
 
+class Genre(models.Model):
```
LiudmylaKulinchenko, line 4: add str method

cinema/serializers.py
LiudmylaKulinchenko, line 1: Why don't you use model serializers instead?

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
LiudmylaKulinchenko: Use get_object_or_404

LiudmylaKulinchenko (changes requested):

cinema/serializers.py
roffi37: How should i know about model serializers at that moment if in video and theory about this theme, said nothing about that?

LiudmylaKulinchenko (approved):
Good job ☃☃

PR #412, opened 2023-11-30
Solution

Wonsky1:


cinema/serializers.py
```diff
@@ -1,24 +1,27 @@
...
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
 
 
-class MovieSerializer(serializers.Serializer):
-    id = serializers.IntegerField(read_only=True)
-    title = serializers.CharField(max_length=255)
-    description = serializers.CharField()
-    duration = serializers.IntegerField()
+class MovieSerializer(serializers.ModelSerializer):
```
MiskoRuslan, line 6: I could be wrong, but it seems that the request was to implement the serializer following the more flexible serializers.Serializer rather than serializers.ModelSerializer. I understand that the functionality in this case is the same in them. However, it would be better to try to implement following serializer.Serializer.

cinema/urls.py
```diff
@@ -1,10 +1,42 @@
...
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
```
MiskoRuslan: maybe it's better to list it first?

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
+    def delete(self, request, pk):
+        genre = self.get_object(pk)
+        genre.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(
+    mixins.ListModelMixin, mixins.CreateModelMixin,
+    generics.GenericAPIView
+):
```
MiskoRuslan, line 73: It may be better to use formatting in this style:
```
class ActorList(
    mixins.ListModelMixin,
    mixins.CreateModelMixin,
    generics.GenericAPIView
):
```

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
+        return self.list(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
+
+
+class ActorDetail(
+    mixins.RetrieveModelMixin, mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin, generics.GenericAPIView
+):
```
MiskoRuslan, line 89: The same, mb better:
```
class ActorDetail(
    mixins.RetrieveModelMixin,
    mixins.UpdateModelMixin,
    mixins.DestroyModelMixin, 
    generics.GenericAPIView,
):
```
but i`m not sure :))

cinema/views.py
```diff
@@ -1,45 +1,114 @@
...
+    def delete(self, request, *args, **kwargs):
+        return self.destroy(request, *args, **kwargs)
+
+
+class CinemaHallViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin, mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin, mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin, generics.GenericAPIView
+):
```
MiskoRuslan, line 114: The same formating 👉👈

MiskoRuslan (changes requested):
🫡

sakhaline (approved):

cinema/serializers.py
sakhaline: in this task it is not necessary to inherit directly from Serializer class, since we are not customizing the functionality, but using the default, so it's good to use ModelSerializer here.

cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+    genres = models.ManyToManyField(Genre, related_name="movies")
 
     def __str__(self):
         return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
```
NEkropo1, line 34: would be a good point to throw a check where we won't let creating halls with negative number of anything.

cinema/serializers.py
```diff
@@ -1,24 +1,27 @@
...
+    class Meta:
+        model = Movie
+        fields = ["title", "description", "duration"]
```
NEkropo1: any other fields?

cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
NEkropo1: better get_object_or_404, since it won't slow down as much as try/except block

NEkropo1 (changes requested):

cinema/models.py
```diff
@@ -1,10 +1,62 @@
...
+
+    def __str__(self) -> str:
+        return self.name
+
+    def clean(self):
+        if self.rows <= 0:
+            raise ValidationError({
+                "row": "Rows must be greater then 0"
```
Y-Havryliv, line 42:
```suggestion
                "row": "Rows must be greater than 0"
```
just a small suggestion

Y-Havryliv (approved):

PR #416, opened 2023-12-01
Solution API Views

Oxbay:


cinema/models.py
```diff
@@ -1,10 +1,107 @@
...
+    actors = models.ManyToManyField(Actor)
+
+    class Meta:
+        ordering = ["title"]
 
     def __str__(self):
         return self.title
+
+
+class MovieSession(models.Model):
```
Y-Havryliv: Seems like these models are from another task

Y-Havryliv (changes requested):

LiudmylaKulinchenko (approved):
Well done ❄

PR #419, opened 2023-12-03
solution

oleksandr-pupchenko:


cinema/admin.py
```diff
@@ -1,8 +1,23 @@
...
+
+
+@admin.register(CinemaHall)
+class CinemaHallAdmin(admin.ModelAdmin):
+    pass
+
+
+@admin.register(Genre)
+class GenreAdmin(admin.ModelAdmin):
+    pass
```
MiskoRuslan, line 23: i think better to use 
```
admin.site.register(model, admin.ModelAdmin)
```

cinema/serializers.py
```diff
@@ -22,3 +27,21 @@ def update(self, instance, validated_data):
         instance.save()
 
         return instance
+
```
MiskoRuslan, line 51: I'm not sure, so feel free to fix it. But it seems to me that you still need to add additional logic to the serializer for movies.

! --  I could be wrong, so feel free to ignore this comment -- !

MiskoRuslan (approved):

cinema/serializers.py
Y-Havryliv: yes, you should handle that

Y-Havryliv (changes requested):

fsocie7y (approved):
GJ 👍

PR #426, opened 2023-12-15
Solution

dzhurarostislav:


cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+    "patch": "partial_update",
+    "delete": "destroy"
+})
 
-from cinema.views import movie_list, movie_detail
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
pavlejviki: When a base path isn't specified, incorporating the router isn't necessary. You can simply define your URL patterns as [list of urls] + router.urls

pavlejviki (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #428, opened 2023-12-21
Solution-py-api-views

Nadiia-developer:


Polyakiv-Andrey (changes requested):
tests not passed!

cinema/migrations/0001_initial.py
```diff
@@ -4,7 +4,6 @@
 
 
 class Migration(migrations.Migration):
-
     initial = True
```
Dimosphen1, line 7: Do not modify the existing migrations.

cinema/models.py
```diff
@@ -1,9 +1,26 @@
...
+    name = models.CharField(max_length=50)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors", blank=True)
+    genres = models.ManyToManyField(Genre)
```
Dimosphen1: The related name is incorrect, you would not get all movies from `Actor` by referring to `actors`. Replace `actors` with `movies` as a related name. You could add a `realated_name` to the `genres` field as well.

cinema/views.py
```diff
@@ -1,45 +1,93 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
Dimosphen1: You could use a `get_object_or_404` shortcut from Django, there is no need to define this method.

Dimosphen1 (changes requested):
Several changes were requested.

cinema/views.py
```diff
@@ -1,45 +1,90 @@
...
+
+    def put(self, request, pk):
+        genre = self.get_object_or_404(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
viktoria-rybenchuk: you can use `is_valid(raise_exception=True)`. 

By default this exception results in a response with the HTTP status code "400 Bad Request". Read documentation https://www.django-rest-framework.org/api-guide/exceptions/

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Well done!

PR #429, opened 2023-12-29
Solution "Solution for api-views"

nsdNite:
Create models, serializers, views, urls.
Update Movie models.


cinema/urls.py
```diff
@@ -1,10 +1,37 @@
...
+             "patch": "partial_update",
+             "delete": "destroy", }
+)
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router when basepath is not provided. Use urlpatterns = [list of urls] + router.urls

cinema/views.py
```diff
@@ -1,45 +1,124 @@
...
+
+        return Response(data=serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "GET":
-        serializer = MovieSerializer(movie)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+    def put(self, request, pk):
+        current_genre = self.get_object(pk)
+        serializer = GenreSerializer(current_genre, data=request.data)
+        if serializer.is_valid():
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #430, opened 2023-12-31
Solution

AnyoneClown:


cinema/views.py
```diff
@@ -1,45 +1,106 @@
...
+        serializer = GenreSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        return get_object_or_404(Genre, pk=pk)
```
Dimosphen1, line 32: It would be better to avoid the creation of methods the single purpose of which is to encapsulate another method. You could use `get_object_or_404` directly where needed so that there would be no need to look into its implementation in other methods.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #439, opened 2024-01-16
Solution

MatsalakO:


viktoria-rybenchuk (changes requested):
All tests should passed

cinema/serializers.py
```diff
@@ -22,3 +22,59 @@ def update(self, instance, validated_data):
...
+
+
+class GenreSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    name = serializers.CharField(max_length=64)
+
+    def create(self, validated_data) -> Genre:
+        return Genre.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
```
pavlejviki: add return type annotation

cinema/urls.py
```diff
@@ -1,10 +1,42 @@
...
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
```
pavlejviki: if you dont use any prefix you can simply append router.urls to your urlpatterns . urlpatterns is a list and router.urls is another list, you can concatenate them using +

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+        serializer = GenreSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
```
pavlejviki: you can use get_object_or_404()

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
```
pavlejviki: when you want to implement a partial update , you can do so by passing partial=True to the serializer. This tells the serializer that the data being passed may contain only a subset of the fields, and it should not raise validation errors for missing fields.

pavlejviki (changes requested):

vasylhnatiuk (approved):

PR #445, opened 2024-01-26
Solution

endernaga:


cinema/urls.py
```diff
@@ -1,10 +1,40 @@
...
+cinema_hall_detail = CinemaHallViewSet.as_view(actions={
+    "get": "retrieve",
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy"
+})
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router when basepath is not provided, use urlpatterns = [list of urls] + router.urls

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
+        serializer = GenreSerializer(genre, data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid():
```

ArsenPidhoretskyi (changes requested):

cinema/urls.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #473, opened 2024-03-03
Solution py-api-views

Vladkrit11:


cinema/urls.py
```diff
@@ -1,10 +1,44 @@
...
+    "get": "retrieve",
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy"
+})
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router when basepath is not provided. Use:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #474, opened 2024-03-05
(py-api-views)

baranovr:
Can you give me any advice on what I should pay attention to in order to pass tests that are failed?

Снимок экрана 2024-03-05 040033


cinema/models.py
```diff
@@ -1,10 +1,34 @@
...
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
 
     def __str__(self):
-        return self.title
+        return f"{self.title}"
```
Dimosphen1, line 34: You could leave the previous version untouched

cinema/views.py
```diff
@@ -1,45 +1,99 @@
...
+):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
+
```
Dimosphen1: There is no need to define methods that are already implemented in mixins. You could remove them. Besides that, place mixins in the fist place in inheritance.

Dimosphen1 (changes requested):
Several changes were requested. Include migration to create required tables so that tests would start to execute

Polyakiv-Andrey (approved):
Good job!

PR #479, opened 2024-03-11
Solution

aLEKS-e3:


MaxymChyncha (approved):
Good job!

cinema/serializers.py
```diff
@@ -22,3 +22,21 @@ def update(self, instance, validated_data):
         instance.save()
 
         return instance
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Actor
+        fields = "__all__"
```
MNDonut, line 30: A small note that between class name and the first subclass/method should be 1 empty line

MNDonut (approved):
There is 1 small one, but in general well done!

PR #489, opened 2024-03-14
Solution

barabarinov:


oksankakot (changes requested):
Your code looks good to me, but fix it so that it passes the check for flake8, 'cause now it all failed because of too long strings

Serhii-Leonenko (approved):
Well done!

PR #490, opened 2024-03-14
solution

mdubyna:


MaxymChyncha (approved):
Good job!

cinema/models.py
```diff
@@ -1,10 +1,43 @@
+from django.contrib.auth.models import AbstractUser
```
MNDonut, line 1: you don't use this import anywhere

cinema/serializers.py
```diff
@@ -1,23 +1,71 @@
...
+    def create(self, validated_data: dict) -> Genre:
+        return Genre.objects.create(**validated_data)
+
+    def update(self, instance: Genre, validated_data: dict) -> Genre:
+        instance.name = validated_data.get("name", instance.name)
+
+        instance.save()
+
+        return instance
+
```
MNDonut, line 42: isn't in better to use ModelSerializer, so that django will do create/update by itself?
Plus you won't need to add these 3 fields and just use Meta class for that

MNDonut (changes requested):

MNDonut (approved):

PR #492, opened 2024-03-14
Solution

AntonSkyrda:


cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request: Request, *args, **kwargs) -> Response:
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request: Request, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
```
tskopenko: U foget to add some logic in this class, coz rn it's not viewing

tskopenko (changes requested):

MNDonut (approved):

PR #498, opened 2024-03-14
Sol

Dobryk12:


cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+cinema_hall_list = CinemaHallViewSet.as_view(
+    actions={"get": "list", "post": "create"}
+)
+cinema_hall_detail = CinemaHallViewSet.as_view(
+    actions={
+        "get": "retrieve",
+        "put": "update",
+        "patch": "partial_update",
+        "delete": "destroy"
+    }
```
isikach, line 25: you should use routers here

isikach (changes requested):

cinema/urls.py
Dobryk12: hz

Serhii-Leonenko (approved):
Well done!

PR #508, opened 2024-03-19
Solution

EugeneProtsenko:


vasylhnatiuk (changes requested):
fix tests

cinema/urls.py
```diff
@@ -1,10 +1,38 @@
...
+        "put": "update",
+        "patch": "partial_update",
+        "delete": "destroy",
+    }
+)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("", include(router.urls)),
```
ArsenPidhoretskyi: No need to include router when basepath is not provided.
Use urlpatterns = [list of urls] + router.urls

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
         if serializer.is_valid():
             serializer.save()
-            return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(serializer.data)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use serializer.is_valid(raise_exception=True) instead of if-else statement

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #528, opened 2024-04-24
Solution

aleksandra-prognimak:


cinema/serializers.py
```diff
@@ -1,13 +1,12 @@
...
 
 
 class MovieSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
-    title = serializers.CharField(max_length=255)
-    description = serializers.CharField()
-    duration = serializers.IntegerField()
+    title = serializers.CharField(max_length=255, required=True)
+    description = serializers.CharField(required=True)
+    duration = serializers.IntegerField(required=True)
```
LiudmylaKulinchenko: They're required by default

LiudmylaKulinchenko: Same everywhere

cinema/urls.py
```diff
@@ -1,10 +1,46 @@
...
+        CinemaHallViewSet.as_view(
+            actions={
+                "get": "retrieve",
+                "put": "update",
+                "patch": "partial_update",
+                "delete": "destroy"
+            }
+        ),
+        name="cinema-hall-detail"
+    ),
```
LiudmylaKulinchenko, line 42: there is no need to specify them when I use default router

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 😸😸

PR #542, opened 2024-05-16
Solution

vsupruniuk:


cinema/models.py
```diff
@@ -1,10 +1,28 @@
...
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
 
+    actors = models.ManyToManyField(to=Actor, related_name="actors")
+    genres = models.ManyToManyField(to=Genre, related_name="genres")
```
LiudmylaKulinchenko: fix related names 👀

LiudmylaKulinchenko (changes requested):
It's ok, but still will be better to add annotations to all your functions/methods!

LiudmylaKulinchenko (approved):
Looks good 😍

PR #556, opened 2024-05-23
class based views DRF

panicua:
I left few comments in ```cinema > urls.py > urlpatterns``` for you easier to understand.
pycharm64_1MfNZqg0Al


cinema/views.py
```diff
@@ -1,45 +1,84 @@
...
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    @staticmethod
+    def get_object(pk: int) -> Genre:
+        return get_object_or_404(Genre, pk=pk)
```
MNDonut: no need to use staticmethod, it works without this decorator, just don't forget to add self back

MNDonut (changes requested):

MNDonut (approved):

PR #564, opened 2024-05-24
Solution

y-kondrashova:


cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    @staticmethod
+    def get_object(pk: int) -> Genre:
+        return get_object_or_404(Genre, pk=pk)
```
MNDonut: no need to use staticmethod, it works without it, just don't forget to add self parameter

MNDonut (changes requested):

MNDonut (approved):

PR #572, opened 2024-05-28
Solution

skeletronys:


cinema/views.py
```diff
@@ -1,45 +1,124 @@
...
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        serializer = GenreSerializer(
+            self.get_object(pk),
+            data=request.data,
+            partial=True,
+        )
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a ValidationError if the data is invalid, which is then caught by the DRF exception handler to return a 400 Bad Request response.

vsmutok (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #579, opened 2024-06-05
Implemented py-api-views

mtiunov:


cinema/views.py
```diff
@@ -1,45 +1,105 @@
...
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+    def post(self, request) -> Response:
+        serializer = GenreSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk:
```suggestion
       serializer.is_valid(raise_exception=True):
```

vasylhnatiuk: you don't need `if` if you use raise_exception=True

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #589, opened 2024-06-16
Solution

AlexRozv:


LiudmylaKulinchenko (approved):
Looks good, but better to add some type annotations! Not to the whole project, but at least to your parts

PR #599, opened 2024-07-02
Solution

Grotski:


cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(
+            self.get_object(pk=pk),
+            data=request.data,
+            partial=True)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(status=status.HTTP_200_OK)
```
vsmutok: you don't need `if` 
better to use  `raise_exception=True` in this case

vsmutok (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        actors = Genre.objects.all()
+        serializer = GenreSerializer(actors, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request) -> Response:
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
             serializer.save()
```
vsmutok: you don't need `if` 
better to use  `raise_exception=True` in this case

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+    def get(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk))
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
         if serializer.is_valid():
             serializer.save()
```
vsmutok: you don't need `if` 
better to use  `raise_exception=True` in this case

vsmutok (changes requested):

cinema/views.py
Grotski: Sorry, my bad I forgot….

EdAlekseiev (approved):

PR #601, opened 2024-07-05
solution

olga-kim7:


cinema/models.py
```diff
@@ -1,9 +1,35 @@
...
+
+    def __str__(self):
+        return self.name
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
viktoria-rybenchuk: add related name

cinema/views.py
```diff
@@ -1,45 +1,108 @@
...
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
viktoria-rybenchuk: [use](https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true) `raise_exception=True` in the `valid` method

viktoria-rybenchuk (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,108 @@
...
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
viktoria-rybenchuk: no need to use if condition

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #604, opened 2024-07-10
solution

boris-devs:


YukihiroSM (changes requested):
FIx tests, please.

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status.HTTP_400_BAD_REQUEST)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
```
vsmutok: the same

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
             serializer.save()
```
vsmutok: the same

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #605, opened 2024-07-14
Solution

VyacheslavFomenko:


cinema/models.py
```diff
@@ -1,10 +1,33 @@
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField(Actor, related_name="actor")
```
vasylhnatiuk: wrong related name

cinema/models.py
```diff
@@ -1,10 +1,33 @@
...
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
vasylhnatiuk: wrong related name

vasylhnatiuk (changes requested):

cinema/models.py
VyacheslavFomenko: why you don't exclude setting.py in flake8, why i should reformate strings that were generated by django ?

vasylhnatiuk (approved):

PR #606, opened 2024-07-19
created serializers and crud views for cinema_hall, genre, actor and …

Timur5050:
…movie


cinema/migrations/0002_actor_cinemahall_genre_movie_actors_movie_genres.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 4.1 on 2024-07-19 09:44
```
pavlopro: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

pavlopro (changes requested):

cinema/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.0.2 on 2022-05-03 13:41
```
pavlopro, line 1: No. You shouldn't modify existing migration. You should create a new one instead. It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

pavlopro (changes requested):

Timur5050:
i dont understand what to do

cinema/migrations/0001_initial.py
pavlopro: Do not modify this file. Please revert changes made here. 
Instead, create a new one migration file and include it to PR

pavlopro (changes requested):

Timur5050:
but there is already one migration and i dont know to to create another one and make only 1 migration

pavlopro (approved):
1. Remove modifications from migration file
2. Represent existing migrations in a new file
Thats all you need to do. But I will approve this.

PR #615, opened 2024-07-25
Solution

YevaYe:


cinema/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
```
maxgerman: import style - see general comments and similar places in code

cinema/models.py
```diff
@@ -8,3 +8,18 @@ class Movie(models.Model):
```
maxgerman, line 10: Movie model should include new fields (see the task)

maxgerman (changes requested):
Good job!
See the comments.
Regarding style - I'd recommend using automated tools for formatting/linting (black, ruff) - these would make the style consistent within the team (and not only) and with minimum diffs

maxgerman (approved):
Nice

PR #616, opened 2024-07-26
Solution

SuskyiVolodymyr:


cinema/serializers.py
```diff
@@ -1,24 +1,36 @@
...
+    class Meta:
+        model = Genre
+        fields = "__all__"
 
-        instance.save()
 
-        return instance
+class MovieSerializer(serializers.ModelSerializer):
+    actors = ActorSerializer(read_only=True, many=True)
+    genres = GenreSerializer(read_only=True, many=True)
```
maxgerman: The task says: "Do not use related serializers for ManyToMany relations."
Think why it's there and what can go wrong (or see the relevant docs section ;)

maxgerman (changes requested):
Good job!
See the comments.
And you shouldn't commit directly to your master/main branch - use the branch as in the task;

maxgerman (approved):

PR #620, opened 2024-07-26
Solution

lilarin:


cinema/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Genre, Actor, CinemaHall
```
maxgerman: imports style - each on new line

cinema/views.py
```diff
@@ -1,45 +1,123 @@
...
+    UpdateModelMixin,
+    DestroyModelMixin
+)
+from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework import status
 
 from django.shortcuts import get_object_or_404
+from rest_framework.views import APIView
+from rest_framework.viewsets import GenericViewSet, ModelViewSet
```
maxgerman: same as above (see other similar cases) and (better) use autoformatter (black etc.)

cinema/views.py
```diff
@@ -1,45 +1,123 @@
...
+    def get(self, request: Request) -> Response:
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request: Request) -> Response:
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
```

cinema/views.py
```diff
@@ -1,45 +1,123 @@
...
+            return Response(serializer.data, status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request: Request, pk: int) -> Response:
+        serializer = GenreSerializer(
+            self.get_object(pk),
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
```
maxgerman: same as above (see similar cases)

maxgerman (changes requested):
Good job! See some comments

DevSpaciX (approved):

PR #621, opened 2024-07-26
solution

DimaYankovskiy:


cinema/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Genre, CinemaHall, Actor
```
maxgerman: imports style - each on new line; use autoformatters e.g. `black`

cinema/views.py
```diff
@@ -1,45 +1,127 @@
...
+    def get(self, request: Request) -> Response:
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request: Request) -> Response:
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
```

maxgerman (changes requested):
Good job!
Consider the comments below

maxgerman (approved):
Well done!

PR #622, opened 2024-07-26
Solution

b3v3kt0r:


cinema/serializers.py
```diff
@@ -22,3 +22,51 @@ def update(self, instance, validated_data):
...
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=255)
+    last_name = serializers.CharField(max_length=255)
+
+    def create(self, validated_data):
+        return Actor.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.first_name = validated_data.get("first_name",
```
maxgerman: this is not a good style - see general comment

cinema/views.py
```diff
@@ -1,45 +1,109 @@
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
-from rest_framework import status
+from rest_framework import status, views, generics, mixins, viewsets
```
maxgerman: style again )

maxgerman (changes requested):
Good job!
Regarding comments - use the automated tools for formatting/linting (black, ruff) - these would make the style consistent within the team (and not only) and with minimum diffs

cinema/serializers.py
```diff
@@ -14,10 +14,10 @@ def create(self, validated_data):
 
     def update(self, instance, validated_data):
         instance.title = validated_data.get("title", instance.title)
-        instance.description = validated_data.get(
-            "description", instance.description
-        )
-        instance.duration = validated_data.get("duration", instance.duration)
+        instance.description = (
+            validated_data.get("description", instance.description))
```
maxgerman, line 18: PEP8 allows several variations on the use of parentheses but the mostly used nowadays and applied by the most popular formatters is smth like this. I'd recommend to stick with it.

```
def example(session):
    result = (
        session.query(models.Customer.id)
        .filter(
            models.Customer.account_id == account_id,
            models.Customer.email == email_address,
        )
        .order_by(models.Customer.id.asc())
        .all()
    )
  ```
  
  See also what the most popular formatter is saying about this:
  https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

File: cinema/views.py [hunk 4]
```diff
@@ -61,9 +60,7 @@ def delete(self, request, pk: int) -> Response:
 
 
 class ActorList(
-    generics.GenericAPIView,
-    mixins.ListModelMixin,
-    mixins.CreateModelMixin
+    generics.GenericAPIView, mixins.ListModelMixin, mixins.CreateModelMixin
```
maxgerman: this was ok, no need to change to one line

maxgerman (changes requested):
added comments

Arsen-hrynevych (approved):
Good job! I really appreciate your effort . Please review the check list of the task.

PR #623, opened 2024-07-27
Develop

BohdanKuzik:


cinema/models.py
```diff
@@ -4,7 +4,33 @@
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField("Actor")
+    genres = models.ManyToManyField("Genre")
```
maxgerman: what if we need to get movies from actors or genres?

cinema/views.py
```diff
@@ -1,45 +1,128 @@
...
+    def get(self, request: Request) -> Response:
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request: Request) -> Response:
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
```
maxgerman: think how we can avoid this "if" and keep the same functionality

cinema/views.py
```diff
@@ -1,45 +1,128 @@
...
-        serializer = MovieSerializer(movie)
+    def get(self, request: Request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk))
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request: Request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk), data=request.data)
         if serializer.is_valid():
```
maxgerman: same as above (and in similar cases in the code)

cinema/serializers.py
```diff
@@ -1,24 +1,35 @@
...
+    class Meta:
+        model = Genre
+        fields = ["id", "name"]
 
-        instance.save()
 
-        return instance
+class MovieSerializer(serializers.ModelSerializer):
+    actors = ActorSerializer(read_only=True, many=True)
+    genres = GenreSerializer(read_only=True, many=True)
```
maxgerman: the task says "Do not use related serializers for ManyToMany relations." Think of other ways to represent these relationships

maxgerman (changes requested):
Well structured code, good!
See some comments.
In general, if your latest migration has some mistake and you notice it, it's better to downgrade and generate it again so as to fix the mistake and avoid additional migration.

maxgerman (approved):
well done!

PR #624, opened 2024-07-27
add solution

jpdreamthug:


cinema/serializers.py
```diff
@@ -1,24 +1,35 @@
...
-        return instance
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = "__all__"
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    actors = ActorSerializer(read_only=True, many=True)
+    genres = GenreSerializer(read_only=True, many=True)
```
maxgerman: The task says: "Do not use related serializers for ManyToMany relations."
Think why it's there and what can go wrong (or see the relevant docs section ;)

cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request: Request, pk: int) -> Response:
+        serializer = GenreSerializer(
+            self.get_object(pk),
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
```

maxgerman (changes requested):
Good job!
The related (esp many-to-many) serializers need some special considerations

maxgerman (approved):
Well done!

PR #625, opened 2024-07-27
Solution

Andrii1000:


cinema/views.py
```diff
@@ -1,45 +1,115 @@
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
-from rest_framework import status
+from rest_framework import status, generics, mixins, viewsets
```
maxgerman: imports style - each on new line

maxgerman (changes requested):
Good job! 
The only nitpick is about style - could be solved by auto formatters (recommended).

cinema/views.py
Andrii1000: have problem with style in each in new line on github don`t pass flake8 tests

DevSpaciX (approved):

PR #627, opened 2024-07-27
'Solution'

Irina17191:


maxgerman (changes requested):
The tests aren't passing :(
Also, migrations are nowhere to be seen in the repository.

Irina17191:
> The tests aren't passing :( Also, migrations are nowhere to be seen in the repository.

I'm sorry, all tests have passed locally in my program, but I didn't add migrations to git and because of that tests don't pass here on git. Fixed it) 

Sorry for my inattention =)

maxgerman (approved):
well done

PR #628, opened 2024-07-28
sol

vladyslavchichkan:


Arsen-hrynevych (changes requested):
Пофікси помилки з тестами.

cinema/models.py
```diff
@@ -1,9 +1,26 @@
...
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
maxgerman: what if we want to get the movies from the actor (reverse relationship)?

cinema/views.py
```diff
@@ -1,45 +1,97 @@
...
             return Response(serializer.data, status=status.HTTP_201_CREATED)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    @staticmethod
```
maxgerman: why staticmethod?

cinema/views.py
```diff
@@ -1,45 +1,97 @@
...
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        serializer = GenreSerializer(self.get_object(pk=pk),
+                                     data=request.data, partial=True)
+        if serializer.is_valid():
```

maxgerman (changes requested):
Good job!
See comments

cinema/views.py
vladyslavchichkan: we can use "try"

cinema/models.py
vladyslavchichkan: we can use related_name

cinema/views.py
vladyslavchichkan: pycharm reccomends to use static here, all the same if we use it or not

Arsen-hrynevych (changes requested):
Unresolved comments from the previous mentor, and I see the possibility that you could format the code for better readability.

Arsen-hrynevych (approved):
Thanks. Try to find a shorter way to write the same functionality.

PR #629, opened 2024-07-28
Solution

dsokolovskiy:


cinema/admin.py
```diff
@@ -1,8 +1,23 @@
 from django.contrib import admin
 
-from cinema.models import Movie
+from cinema.models import Movie, Genre, Actor, CinemaHall
```
maxgerman, line 3: multiple imports - style

cinema/views.py
```diff
@@ -1,45 +1,113 @@
...
-    if request.method == "GET":
-        movies = Movie.objects.all()
-        serializer = MovieSerializer(movies, many=True)
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+
+class GenreList(APIView):
+    @staticmethod
```
maxgerman, line 22: why staticmethod?
I understand it could be because of the IDE suggestions but it's critical to understand the meaning of this.
Try experimenting and using `request` obj in this method to find out :)

maxgerman (changes requested):
Good job!
Some nitpicks about style - could be solved by auto formatters (recommended).
Pay attention to these:
- you shouldn't commit directly to your master/main branch - use the branch as in the task;
- revise the usual methods, classmethods, staticmethods

PR #630, opened 2024-07-29
Solution

Yuruch:


cinema/views.py
```diff
@@ -1,45 +1,129 @@
...
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request: Request, pk: int) -> Response:
+        serializer = GenreSerializer(
+            self.get_object(pk),
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
```
maxgerman: how can we avoid this and similar "if" statements and save the same functionality?

maxgerman (changes requested):
Great job!
The only recommended thing is not very significant but can make the code more concise, less nested

DevSpaciX (approved):

PR #633, opened 2024-07-31
Solution

dsokolovskiy:


cinema/views.py
```diff
@@ -1,45 +1,115 @@
...
-            serializer.save()
-            return Response(serializer.data, status=status.HTTP_201_CREATED)
+class GenreList(APIView):
+    def get(self, request) -> Response:
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    @staticmethod
```
maxgerman: why staticmethod? (here and in other similar places)

cinema/views.py
```diff
@@ -1,45 +1,115 @@
-from rest_framework.decorators import api_view
 from rest_framework.response import Response
-from rest_framework import status
+from rest_framework import status, generics, mixins, viewsets
```
maxgerman: imports style - each on new line

maxgerman (changes requested):
Good job!
Revise staticmethods and usual methods.
Small style nitpicks - could be solved by tools for auto formatting/linting (black, ruff) - these would make the style consistent within the team (and not only) and with minimum diffs

DevSpaciX (approved):

PR #639, opened 2024-08-08
Solution

OLENA-KALITSINSKA:


cinema/serializers.py
```diff
@@ -22,3 +22,66 @@ def update(self, instance, validated_data):
...
+        instance.first_name = validated_data.get(
+            "first_name",
+            instance.first_name
+        )
+        instance.last_name = validated_data.get(
+            "last_name",
+            instance.last_name
+        )
+        instance.save()
+        return instance
```
Dimosphen1: You could inherit `ActorSerializer` from `serializers.ModelSerializer` and in this way avoid defining `create` and `update` methods. Make this here and in the other serializers as well

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
-        movie.delete()
+    def delete(self, request, pk: int) -> Response:
+        self.get_object(pk=pk).delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(
+    generics.GenericAPIView,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
```
Dimosphen1: It is a good practice to place mixins first in inheritance before other classes. Change the inheritance order here and in the other views as well

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs) -> Response:
+        return self.update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs) -> Response:
+        return self.partial_update(request, *args, **kwargs)
```
Dimosphen1, line 92: There is no need to add these methods since they are already included in the mixins. Remove them here and in the other views as well

Dimosphen1 (changes requested):
Several changes were requested

cinema/serializers.py
```diff
@@ -22,3 +22,66 @@ def update(self, instance, validated_data):
...
+
```
pavlopro, line 31: You could inherit ActorSerializer from serializers.ModelSerializer and in this way avoid defining create and update methods. Make this here and in the other serializers as well

cinema/views.py
pavlopro: This was not resolved

pavlopro (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #640, opened 2024-08-09
Solution

MykhailoIvchenko:


cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
-        if serializer.is_valid():
+    def put(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```

cinema/serializers.py
```diff
@@ -1,6 +1,57 @@
...
+        return Actor.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.first_name = validated_data.get("first_name",
+                                                 instance.first_name)
+        instance.last_name = validated_data.get("last_name",
+                                                instance.last_name)
+        instance.save()
+
+        return instance
```

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs) -> Response:
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs) -> Response:
+        return super().partial_update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
```
vsmutok, line 91: There is no need to add these methods since they are already included in the mixins. Remove them here and in the other views as well

vsmutok (changes requested):

cinema/views.py
MykhailoIvchenko: Done

cinema/serializers.py
MykhailoIvchenko: Done

cinema/views.py
MykhailoIvchenko: I can't do that, the classes for Actor inherit GenericAPIView, not GenericViewSet (according to the task). If I remove the methods, there are tests failures.

cinema/models.py
```diff
@@ -1,10 +1,27 @@
...
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
vsmutok: add related name

cinema/views.py
```diff
@@ -1,45 +1,106 @@
...
```
vsmutok, line 91:
```suggestion
class ActorList(generics.ListCreateAPIView):
    queryset = Actor.objects.all()
    serializer_class = ActorSerializer


class ActorDetail(generics.RetrieveUpdateDestroyAPIView):
    queryset = Actor.objects.all()
    serializer_class = ActorSerializer
```

vsmutok (changes requested):

cinema/models.py
MykhailoIvchenko: Done

cinema/views.py
MykhailoIvchenko: Task
I also don't like the existing solution, but there is the task where we can see the strict instruction "For the Actor model use a GenericAPIView" (not ListCreateAPIView and RetrieveUpdateDestroyAPIView).

vsmutok: ok, if the task condition calls for it let's leave it at that. But in the future, use that example as I recommend

vsmutok (approved):
Great!

PR #641, opened 2024-08-11
сreate Genre Actor CinemaHall models & update Movie

SerheiZhus:


cinema/serializers.py
```diff
@@ -1,24 +1,95 @@
...
+        instance.first_name = validated_data.get(
+            "first_name",
+            instance.first_name
+        )
+        instance.last_name = validated_data.get(
+            "last_name",
+            instance.last_name
+        )
+        instance.save()
+        return instance
```

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs) -> Response:
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs) -> Response:
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
```
vsmutok: There is no need to add these methods since they are already included in the mixins. Remove them here and in the other views as well

vsmutok (changes requested):

EdAlekseiev (approved):

PR #642, opened 2024-08-12
Solution

Meksim-fesh:


cinema/serializers.py
```diff
@@ -22,3 +39,65 @@ def update(self, instance, validated_data):
...
+        instance.first_name = validated_data.get(
+            "first_name", instance.first_name
+        )
+        instance.last_name = validated_data.get(
+            "last_name", instance.last_name
+        )
+
+        instance.save()
+
+        return instance
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #643, opened 2024-08-13
Solution

IraMladanovych02:


cinema/migrations/0003_alter_actor_first_name_alter_actor_last_name_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+    def delete(self, request, pk):
+        genre = self.get_object(pk)
+        genre.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(
+    generics.GenericAPIView,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
```
Dimosphen1: Place mixins on a first position in inheritance before generics here and in the other places as well

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs):
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        return self.destroy(request, *args, **kwargs)
```
Dimosphen1: There is no need to create these methods here and in the other places, mixins already provide this functionality. Remove them

Dimosphen1 (changes requested):
Several changes were requested

cinema/serializers.py
```diff
@@ -1,13 +1,71 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=64)
+    last_name = serializers.CharField(max_length=64)
```
LLkaia: What will be if Serializer will allow length 64 to be written to database, and then Model has 55 length allowed?

cinema/views.py
```diff
@@ -1,45 +1,87 @@
...
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+class GenreDetail(APIView):
+    def get_object(self, pk):
```
LLkaia: This method can be static

LLkaia (changes requested):
Few adjustments

LLkaia (approved):
GJ!

PR #658, opened 2024-09-15
Solution

raychw:


cinema/migrations/0001_initial.py
Dimosphen1, line 1: Do not redefine an existing migration, create a new one instead

cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
+    def get(self, request):
+        genre = Genre.objects.all()
+        serializer = GenreSerializer(genre, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
```
Dimosphen1: Use `raise_exception=True` here and in the other places as well to remove the if on condition while validating serializer

Dimosphen1 (changes requested):
Make sure that all checks pass. Several changes were requested

cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
vsmutok: 1. Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

Good example:
```python
if request.method == 'POST':
    serializer.save()
```

Bad example:
```python
if request.method == 'POST':
    if serializer.is_valid():
        serializer.save()
```

vsmutok (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #661, opened 2024-09-20
implemented views via class based views - APIView, GenericAPIView, Ge…

MasakDirt:
…nericViewSet, ModelViewSet and routers


cinema/serializers.py
```diff
@@ -1,13 +1,83 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=100, required=True)
```
dimak20, line 8: You use both variants: required=True and sometimes without required statement. I think you should choose what you want to use

cinema/serializers.py
```diff
@@ -1,13 +1,83 @@
...
+
+        return instance
 
 
 class MovieSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
     title = serializers.CharField(max_length=255)
     description = serializers.CharField()
     duration = serializers.IntegerField()
+    actors = serializers.SlugRelatedField(
```
dimak20, line 71: Is It required to create full_name in README.md?

cinema/serializers.py
```diff
@@ -1,13 +1,83 @@
...
+        many=True,
+        read_only=True,
+        slug_field="full_name"
+    )
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
```
dimak20, line 78: Would It possible to create Movie and add genres simultaneously with this one serializer? Provided you use just one seriaizer to display list and create movie I don't think so

dimak20 (approved):

oliasymak24 (approved):

PR #662, opened 2024-09-21
'Solution'

oleksashcherbakov:
image
image
image
image
image
image
image


cinema/views.py
```diff
@@ -1,45 +1,86 @@
...
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #664, opened 2024-09-22
solution

BastovOleksandr:


oliasymak24 (changes requested):
Fix annotations everywhere

cinema/views.py
```diff
@@ -1,45 +1,107 @@
...
+        genre.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def patch(self, request: Request, pk: int):
+        serializer = GenreSerializer(
+            self.get_object(pk), data=request.data, partial=True
+        )
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
```
Arsen-hrynevych, line 53: In GenreDetail, you're manually implementing get, put, delete, and patch. Since DRF has built-in generic views (RetrieveUpdateDestroyAPIView), this can be avoided.

Arsen-hrynevych (approved):
Consider simplifying your code, as there are unnecessary steps that can be eliminated.

PR #666, opened 2024-09-25
+

AndrZhuk:


cinema/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.0.2 on 2022-05-03 13:41
+# Generated by Django 4.1 on 2024-09-25 16:28
```
pavlopro: Do not modify existing migration. Create a new one instead

cinema/models.py
```diff
@@ -1,9 +1,26 @@
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
pavlopro: Add related names for relational fields

cinema/views.py
```diff
@@ -1,45 +1,85 @@
...
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
```
pavlopro: Use `get_object_or_404 `instead

pavlopro (changes requested):

cinema/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.0.2 on 2022-05-03 13:41
+# Generated by Django 4.1 on 2024-09-25 18:17
```
pavlopro: Do not modify existing migration file. Create a new one instead

cinema/models.py
```diff
@@ -1,9 +1,26 @@
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movie")
+    genres = models.ManyToManyField(Genre, related_name="movie")
```
pavlopro: Not correct related names. Should be `movies`

pavlopro (changes requested):

cinema/migrations/0001_initial.py
AndrZhuk: I am not change old file, I am delete old and create new

AndrZhuk: I don't understand this comment at all, if there is more than 1 migration file, you write delete it and make only 1, now you say don't change this first file

AndrZhuk: I guess I don't understand the mentors' policy regarding the migration file, everyone will always have their own new opinion that will be the reason for rejecting the project

AndrZhuk: I'm really exhausted in these comments, please explain to me why it is a good practice to have 2 migration files for this project, and for some other project, there should be only one

EdAlekseiev (approved):

PR #668, opened 2024-09-26
Solution

dartomOOv:


cinema/serializers.py
```diff
@@ -22,3 +22,57 @@ def update(self, instance, validated_data):
...
+    def update(self, instance, validated_data):
+        instance.first_name = validated_data.get(
+            "first_name", instance.first_name
+        )
+        instance.last_name = validated_data.get(
+            "last_name", instance.last_name
+        )
+        instance.save()
+
+        return instance
```
DSahalatyi, line 45: I am not entirely sure if this is required, but how would one add genres and/or actors to the movie when creating/updating?

cinema/serializers.py
```diff
@@ -22,3 +22,57 @@ def update(self, instance, validated_data):
...
+
+
+class GenreSerializer(serializers.Serializer):
```
DSahalatyi, line 48: You don't want to rely on DB errors for unique constraint validation.

DSahalatyi (changes requested):

cinema/serializers.py
dartomOOv: It isn't required,
зображення
and tests passing without errors

cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+        CinemaHallViewSet.as_view(
+            actions={
+                "get": "retrieve",
+                "put": "update",
+                "patch": "partial_update",
+                "delete": "destroy",
+            }
+        ),
+        name="cinema-hall-detail"
+    ),
```
DSahalatyi, line 39: Assigning views configurations to variables and then using these variables in the urlpatterns can give you a much cleaner overall look of the url patterns.

DSahalatyi (approved):

oliasymak24 (changes requested):
Fix annotations everywhere

oliasymak24 (approved):

PR #670, opened 2024-09-26
first commit

panichqa:


cinema/views.py
```diff
@@ -1,45 +1,120 @@
...
+        self.get_object(pk=pk).delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def patch(self, request, pk) -> Response:
+        genre = self.get_object(pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        serializer.is_valid()
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
```
dartomOOv: What if serializer is not valid, where errors will go as in 'put' method?

(optional). You can also reduce the amount of code here :)

dartomOOv (changes requested):

dartomOOv (approved):

Arsen-hrynevych (approved):
overall ok,  but it could be made simpler.

PR #672, opened 2024-09-26
first_try

yashkunn:


cinema/models.py
```diff
@@ -1,10 +1,40 @@
...
+
+    def __str__(self):
+        return self.name
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    genres = models.ManyToManyField(Genre)
```
Sergi0bbb: is bad practise to not set releted_name

cinema/views.py
```diff
@@ -1,45 +1,117 @@
...
-
+            return Response(serializer.data)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    if request.method == "DELETE":
-        movie.delete()
+    def patch(self, request, pk) -> Response:
+        genre = self.get_object(pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        serializer.is_valid()
```
Sergi0bbb: what would be here if we get invalid data?

Sergi0bbb (changes requested):

Arsen-hrynevych (approved):
Ok

PR #677, opened 2024-09-27
Develop

Narberal90:


cinema/serializers.py
```diff
@@ -1,23 +1,76 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
 
 
-class MovieSerializer(serializers.Serializer):
+class MovieSerializer(serializers.ModelSerializer):
```
mnezbrytska: If you want to update a film by adding only one actor to the existing ones, can you do it without any additional logic here?

mnezbrytska (approved):

Arsen-hrynevych (approved):
Overall ok,  but it could be made simpler.

PR #678, opened 2024-09-27
Develop

vladislav-tsybuliak1:


Arsen-hrynevych (approved):
Overall ok, but it could be made simpler.

PR #680, opened 2024-09-27
Solution Politov

Alexandr-Politov:


cinema/serializers.py
```diff
@@ -1,24 +1,94 @@
...
+from rest_framework.validators import UniqueValidator
 
-from cinema.models import Movie
+from cinema.models import Actor, Genre, CinemaHall, Movie
+
+
+class ActorSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    first_name = serializers.CharField(max_length=50)
+    last_name = serializers.CharField(max_length=50)
```
BastovOleksandr, line 10: I'm not sure, but I think it is time to use serializers.ModelSerializer instead of Serializer
so your "serializers.py" code will become much shorter
4 strings almost for each class, except Movie with it ManyToMany realations

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
 
-@api_view(["GET", "POST"])
-def movie_list(request):
-    if request.method == "GET":
-        movies = Movie.objects.all()
-        serializer = MovieSerializer(movies, many=True)
+class GenreList(views.APIView):
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
```
BastovOleksandr: I got a reject for not using parameter type annotations, including what the method returns

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
+    def delete(self, request, pk: int):
+        genre = self.get_object(pk=pk)
+        genre.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(
+    generics.GenericAPIView,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
```
BastovOleksandr: order is matter
mixins are going first

cinema/views.py
```diff
@@ -1,45 +1,109 @@
...
+        return self.list(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
+
+
+class ActorDetail(
+    generics.GenericAPIView,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
```
BastovOleksandr: same here, just fix it everywhere

BastovOleksandr (changes requested):
I double checked, it seems like it's time to use serializers.ModelSerializer

BastovOleksandr (approved):

cinema/serializers.py
```diff
@@ -1,24 +1,70 @@
...
+#         )
+#         instance.duration = validated_data.get("duration", instance.duration)
+#         if "actors" in validated_data:
+#             actors = validated_data.pop("actors")
+#             instance.actors.set(actors)
+#         if "genres" in validated_data:
+#             genres = validated_data.pop("genres")
+#             instance.genres.set(genres)
+#         instance.save()
+#         return instance
```
Arsen-hrynevych: Commented-out code should not be included in the repository.

cinema/urls.py
```diff
@@ -1,10 +1,42 @@
...
-from cinema.views import movie_list, movie_detail
+from cinema.views import (
+    GenreList,
+    GenreDetail,
+    ActorList,
+    ActorDetail,
+    CinemaHallViewSet,
+    MovieViewSet,
+)
+
```
Arsen-hrynevych, line 12: 2 blank lines after imports

cinema/views.py
```diff
@@ -1,45 +1,107 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request: Request, *args, **kwargs) -> Response:
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request: Request, *args, **kwargs) -> Response:
+        return self.update(request, partial=True, *args, **kwargs)
+
+    def delete(self, request: Request, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
```
Arsen-hrynevych, line 90: Mixins already take care of this for you, making the code redundant.

Arsen-hrynevych (changes requested):

cinema/views.py
Alexandr-Politov: No. Mixins take care of this methods only with GenericViewSet. With GenericAPIView they only let you use this simplified srtucture.
image

cinema/serializers.py
```diff
@@ -1,24 +1,35 @@
...
+class MovieSerializer(serializers.ModelSerializer):
+    actors = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=Actor.objects.all()
+    )
+    genres = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=Genre.objects.all()
+    )
+
+    class Meta:
+        model = Movie
```
MNDonut: We'll learn these serializers in API Serilizer topic, for this you should still use Serializer because
image

cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+from cinema.views import (
+    GenreList,
+    GenreDetail,
+    ActorList,
+    ActorDetail,
+    CinemaHallViewSet,
+    MovieViewSet,
+)
+
```
MNDonut, line 12: 2 blank lines after the last import

cinema/views.py
MNDonut: @Alexandr-Politov actually for both generic classes you should use mixins(you are correct!). Mixins define actions (list, create, destroy,..) why Viewset and View have (get, post, delete,...) methods. But they aren't connected. By default DRF doesn't know that for "def post" it should call "create" action. This is why we need to do it explicitly.
Or we can inherit from ModelViewset/ListCreateAPIView/RetrieveUpdateDestroyView that already have connected methods with actions. The best idea it to view DRF code by opening these classes and check their implementation ;)

MNDonut (changes requested):

cinema/views.py
```diff
@@ -1,45 +1,107 @@
...
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
-            serializer.save()
-            return Response(serializer.data, status=status.HTTP_201_CREATED)
+    def post(self, request: Request) -> Response:
+        serializer = GenreSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
Oleksl888, line 26: In this task it actually asks to 

> Don't forget return Response with errors if serializer is not valid:

Oleksl888 (approved):
Good job!

PR #681, opened 2024-09-27
Solution

Poznanskyi:


dartomOOv (changes requested):
Also, make test work before pushing solution

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+    def post(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.create(request, *args, **kwargs)
+
+
+class ActorDetail(
+    generics.GenericAPIView,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+):
```
dartomOOv, line 79: Will the patch method work if there is no needed mixin inherited?

cinema/views.py
```diff
@@ -1,45 +1,110 @@
...
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
+
+
+class CinemaHallViewSet(
```
dartomOOv, line 96: Same here.

cinema/views.py
```diff
@@ -1,45 +1,110 @@
-from rest_framework.decorators import api_view
+from django.core.serializers import serialize
```
dartomOOv: Unused import I suppose.

cinema/views.py
dartomOOv, line 78: Will the patch method work if there is no needed mixin inherited?

cinema/views.py
dartomOOv, line 95: Same here.

cinema/views.py
```diff
@@ -1,45 +1,112 @@
...
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+    def patch(self, request, *args, **kwargs):
+        return self.partial_update(request, *args, **kwargs)
```
dartomOOv, line 107: You can remove this.

dartomOOv (approved):
I`m very sorry, I forgot that 'partial_update' func is in the same mixin as 'update' func.

cinema/urls.py
```diff
@@ -1,10 +1,27 @@
-from django.urls import path
+from django.urls import path, include
+from rest_framework import routers
```
MNDonut, line 2: split different groups of imports

MNDonut (approved):

PR #683, opened 2024-09-27
Solution

OleksiukStepan:


HalynaRohatska (approved):

Arsen-hrynevych (changes requested):

cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}: rows - {self.rows}, seats - {self.seats_in_row}"
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
```
Arsen-hrynevych: Using IntegerField for movie duration might work, but it's unclear what unit of time you're storing (minutes, seconds, etc.). It might be beneficial to add a help_text attribute or use DurationField for clarity and flexibility.

cinema/serializers.py
```diff
@@ -1,24 +1,63 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
```
Arsen-hrynevych: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

cinema/urls.py
```diff
@@ -1,10 +1,48 @@
...
-from cinema.views import movie_list, movie_detail
+from cinema.views import (
+    MovieViewSet,
+    GenreList,
+    GenreDetail,
+    ActorList,
+    ActorDetail,
+    CinemaHallViewSet,
+)
+
```
Arsen-hrynevych, line 12: 2 blank lines after imports

cinema/serializers.py
```diff
@@ -1,24 +1,63 @@
...
 
 
 class MovieSerializer(serializers.Serializer):
```
Arsen-hrynevych: The MovieSerializer could be simplified by using ModelSerializer instead of Serializer. Since you're working with a Django model, ModelSerializer automatically handles common functionality (like field definitions and create/update methods) for you.

cinema/serializers.py
```diff
@@ -1,24 +1,63 @@
...
-    def update(self, instance, validated_data):
+    def update(self, instance: Movie, validated_data: dict) -> Movie:
         instance.title = validated_data.get("title", instance.title)
         instance.description = validated_data.get(
             "description", instance.description
         )
         instance.duration = validated_data.get("duration", instance.duration)
+        actors = validated_data.get("actors", None)
+        genres = validated_data.get("genres", None)
+
```
Arsen-hrynevych: In the update method of MovieSerializer, the checks for actors and genres could be simplified. Instead of checking whether they exist and updating only if provided, you might directly set these values, as .set() works with empty lists as well.

OleksiukStepan:
> You've duplicated a lot of code that's already implemented in Django. To adhere to the DRY (Don't Repeat Yourself) principle, you could simplify your code.

Could you please provide more details on what I need to refactor? I have seen another approved solution for this task and identified two areas for improvement:
1. change
genre = self.get_object(pk)
genre.delete()
to 
genre.self.get_objects(pk).delete()

2. Use ModelSerializer for MovieSerializer (what as I understand unnecessary for this task)

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+    generics.GenericAPIView
+):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+    def get(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.list(request, *args, **kwargs)
+
+    def post(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.create(request, *args, **kwargs)
```
Arsen-hrynevych: In your ActorList and ActorDetail classes, the get and post methods could be simplified further by calling the methods directly.

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request: HttpRequest, *args, **kwargs) -> Response:
+        return self.destroy(request, *args, **kwargs)
```
Arsen-hrynevych: No need for overrides:
Since ListModelMixin provides the logic for listing and creating objects, and RetrieveModelMixin, UpdateModelMixin, and DestroyModelMixin handle retrieval, updating, and deletion, you don't need to redefine those methods.

Arsen-hrynevych (changes requested):
Thank you! Instead of just resending your task for review, please take a moment to look it over or ask any questions. After making the fixes, I'll wait for the solution.

cinema/views.py
OleksiukStepan: I cannot delete these methods because it will cause test failures in both classes

OleksiukStepan: OK. I've found the solution for this already. I have changed inherit, but in READMY.md write that for the Actor model I need to use a GenericAPIView

oliasymak24 (approved):

PR #688, opened 2024-09-28
solution

HalynaRohatska:


pashawarganov (approved):
Code looks good. I think your problem in migrations

PR #696, opened 2024-10-01
solution

HalynaRohatska:


cinema/serializers.py
```diff
@@ -1,16 +1,47 @@
...
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = ["name", "rows", "seats_in_row"]
 
 
 class MovieSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
     title = serializers.CharField(max_length=255)
     description = serializers.CharField()
```
BastovOleksandr: You don't need to write all these fields in MovieSerializer
just inherit your serializer from ModelSerializer, like you did before. It should work
The only thing you have to keep is the ManyToMany fields, that's all
No other fields, no methods overriding

cinema/views.py
```diff
@@ -1,45 +1,86 @@
...
 
-    if request.method == "POST":
-        serializer = MovieSerializer(data=request.data)
+    def post(self, request) -> Response:
+        serializer = GenreSerializer(data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_201_CREATED)
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
BastovOleksandr: you can do without the "if" statement and the redundant "return" statement by simply using the "raise_exception=True" key as an argument to the is_valid() method

cinema/views.py
```diff
@@ -1,45 +1,86 @@
...
+        return Response(serialiser.data, status=status.HTTP_200_OK)
 
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
+    def put(self, request, pk: int) -> Response:
+        serializer = GenreSerializer(self.get_object(pk=pk), data=request.data)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
BastovOleksandr: same here, just fix it everywhere. Its easy)

cinema/views.py
```diff
@@ -1,45 +1,86 @@
...
+
+
+class ActorList(generics.ListCreateAPIView):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class ActorDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
```
BastovOleksandr: I'm not sure if we can use "ListCreateAPIView" and "ListCreateAPIView" in the solution because the task says to use "GenericAPIView" with needed mixins and methods (get, put, delete, patch, post)
I think this was invented so that we could just understand a little more deeply how it works. Technically, your solution is correct, of course.

BastovOleksandr (changes requested):
one of the comments (you will understand which one from the context) to correct at your own discretion

cinema/serializers.py
```diff
@@ -35,6 +31,17 @@ class MovieSerializer(serializers.Serializer):
...
+            "title",
+            "description",
+            "duration",
+            "actors",
+            "genres"
+        ]
+
     def create(self, validated_data):
         actors = validated_data.pop("actors")
         genres = validated_data.pop("genres")
```
BastovOleksandr, line 47: i think you can delete it

BastovOleksandr (approved):

Arsen-hrynevych (approved):
Overall ok

PR #699, opened 2024-10-06
solution

AndreiVed:


cinema/views.py
```diff
@@ -1,45 +1,118 @@
...
-            return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(serializer.data)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def patch(self, request, pk):
+        genre = self.get_object(pk=pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
```

vsmutok (changes requested):

vsmutok (approved):
Looks Good!

PR #723, opened 2024-11-23
Solution

vladyslav-tmf:


cinema/serializers.py
```diff
@@ -1,24 +1,73 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Actor, CinemaHall, Genre, MAX_LENGTH, Movie
```
Arsen-hrynevych, line 3: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

cinema/urls.py
```diff
@@ -1,10 +1,40 @@
...
+    path("actors/", views.ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", views.ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls))
+]
```
Arsen-hrynevych, line 40: The style isn't consistent

Arsen-hrynevych (approved):
ok

PR #728, opened 2024-11-29
Solution

VladyslavBon:


cinema/serializers.py
```diff
@@ -1,24 +1,27 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
```
Arsen-hrynevych: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

cinema/urls.py
```diff
@@ -1,10 +1,43 @@
...
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("cinema_halls/", cinema_hall_list, name="cinema-hall-list"),
+    path(
+        "cinema_halls/<int:pk>/",
+        cinema_hall_detail,
+        name="cinema-hall-detail"
+    ),
+    path("", include(router.urls)),
+
```
Arsen-hrynevych: It could be simplified.

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs):
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        return self.destroy(request, *args, **kwargs)
```
Arsen-hrynevych: Right now, you are violating the DRY principle, which means you have done the same thing that has already been implemented in the classes you are using with mixins

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+    generics.GenericAPIView
+):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
```
Arsen-hrynevych: Right now, you are violating the DRY principle, which means you have done the same thing that has already been implemented in the classes you are using with mixins

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
             return Response(serializer.data, status=status.HTTP_201_CREATED)
-
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
```
Arsen-hrynevych: I don't see the point of creating a helper method that has already been implemented in Django. I don't see any advantage or usefulness added by using a custom function.

cinema/views.py
```diff
@@ -1,45 +1,111 @@
...
+from rest_framework.views import APIView
 
-from django.shortcuts import get_object_or_404
+from cinema.models import Movie, Actor, Genre, CinemaHall
+from cinema.serializers import (
+    MovieSerializer,
+    ActorSerializer,
+    GenreSerializer,
+    CinemaHallSerializer
+)
```
Arsen-hrynevych: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych: Also, format the imports because they do not follow alphabetical order.

Arsen-hrynevych (changes requested):
Mostly correct. Keep going!

cinema/views.py
Arsen-hrynevych: Not fixed

cinema/views.py
```diff
@@ -1,45 +1,116 @@
...
+        return self.retrieve(request, *args, **kwargs)
+
+    def put(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs):
+        return self.partial_update(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        return self.destroy(request, *args, **kwargs)
```
Arsen-hrynevych: The same problem here

Arsen-hrynevych (changes requested):
Please fix DRY violation

Arsen-hrynevych (approved):
ok

PR #760, opened 2024-12-10
Solution

yulia-pl:


cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+
+    def __str__(self):
+        return self.name
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
+    actors = models.ManyToManyField(Actor)
```
Bondzik-S: Add `related_name` for ManyToManyField

cinema/models.py
```diff
@@ -1,10 +1,36 @@
...
+    genres = models.ManyToManyField(Genre)
```
Bondzik-S: The same here

Bondzik-S (changes requested):

cinema/models.py
yulia-pl: Thank you very much 🙏🙏🙏

PR #800, opened 2025-02-09
Solution

julia4406:


cinema/views.py
```diff
@@ -1,45 +1,124 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            raise Http404
```
Dogober: herer you can use such method
`    def get_object(self, pk: int) -> Genre:
        return get_object_or_404(Genre, pk=pk)`

Dogober (approved):
Good job

PR #806, opened 2025-02-10
solution-for-api-views

Bleit1:
изображение


cinema/serializers.py
```diff
@@ -1,24 +1,64 @@
...
+        fields = ["id", "name"]
+
+
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = ["id", "name", "rows", "seats_in_row"]
+
+
+class MovieSerializer(serializers.ModelSerializer):
```
ruslanyemets, line 23: Try leaving the default values ​​in the MovieSerializer class to pass the tests (without actors, genres, pop and set methods, etc.)

ruslanyemets (approved):
Good job!

PR #819, opened 2025-02-11
Unfinished

bodiakof:


cinema/serializers.py
```diff
@@ -9,15 +9,78 @@ class MovieSerializer(serializers.Serializer):
...
+        instance.duration = validated_data.get(
+            "duration",
+            instance.duration
+        )
+        instance.save()
+
+        return instance
+
+
+class GenreSerializer(serializers.Serializer):
```
Nattalli: why don't you use `ModelSerializer`?
you can them don't overwrite `create` and `update` methods

Nattalli (changes requested):

cinema/serializers.py
bodiakof: Yeah, I know about it. It's just [here](https://github.com/bodiakof/py-movie-api/blob/dev/cinema/serializers.py) where I guess, I was actually expected to use regular Serializer, I used ModelSerializer instead, so I figured why not use Serializer here, since I kinda cheated in that task :) Do you want me to change the code with ModelSerializer?

Nattalli: yes, it's better to change. You have to make a decision depending on the situation. in this case, it's better to use ModelSerializer, because here models are really used + there is no custom logic for fields + then you don't have to redefine methods unnecessarily

Nattalli (changes requested):
good but use ModelSerializer

Nattalli (approved):

PR #823, opened 2025-02-13
Solution

serg121192:


cinema/views.py
```diff
@@ -1,45 +1,129 @@
...
+            )
 
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    def patch(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data, partial=True)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        else:
```

cinema/views.py
```diff
@@ -1,45 +1,129 @@
...
-    if request.method == "PUT":
-        serializer = MovieSerializer(movie, data=request.data)
-        if serializer.is_valid():
+    def put(self, request, pk):
+        genre = self.get_object(pk)
+        serializer = GenreSerializer(genre, data=request.data)
+        if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
+        else:
```
vsmutok: 1. Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

Good example:
```python
if request.method == 'POST':
    serializer.save()
```

Bad example:
```python
if request.method == 'POST':
    if serializer.is_valid():
        serializer.save()
```

vsmutok (changes requested):

SerhiyTo (approved):

PR #838, opened 2025-03-15
Solution

Wenhun:


Wenhun:
@mateacademy-ai-mentor I do not agree with the proposed changes

PR #854, opened 2025-04-06
Develop

Leo9siy:


cinema/views.py
```diff
@@ -43,3 +46,110 @@ def movie_detail(request, pk):
...
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```

vsmutok (changes requested):

cinema/views.py
Leo9siy: ok

cinema/views.py
```diff
@@ -34,12 +35,115 @@ def movie_detail(request, pk):
...
+    def get(self, request):
+        genres = Genre.objects.all()
+        serializer = GenreSerializer(genres, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def post(self, request):
+        serializer = GenreSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```


Good example:
```python
if request.method == 'POST':
    serializer.save()
```

Bad example:
```python
if request.method == 'POST':
    if serializer.is_valid():
        serializer.save()
```

vsmutok (changes requested):

cinema/views.py
Leo9siy: Ok, now i got it

vsmutok (approved):
Well Done!

PR #859, opened 2025-04-08
Solution

Baranotik15:


cinema/urls.py
```diff
@@ -1,10 +1,31 @@
...
+
+router.register("actor", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genre", GenreViewSet)
 
 urlpatterns = [
-    path("movies/", movie_list, name="movie-list"),
-    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+    path("movies/", movie_list),
+    path("movies/<int:pk>/", movie_detail),
```
Boyok1337: Missed name for path

cinema/urls.py
```diff
@@ -1,10 +1,31 @@
...
+    path("actors/", ActorList.as_view(), name="actor-list"),
+    path("actors/<int:pk>/", ActorDetail.as_view(), name="actor-detail"),
+    path("genres/", GenreList.as_view(), name="genre-list"),
+    path("genres/<int:pk>/", GenreDetail.as_view(), name="genre-detail"),
```
Boyok1337: Why do you need the same paths in both routers and paths?

Boyok1337 (changes requested):

cinema/urls.py
TopalRuslan: confused by using two approaches in view

cinema/urls.py
TopalRuslan: Do we need to use names if we want to set the same ones as the default ones?

Boyok1337: 1. Explicit is better than implicit
2. Do we have default names for func-based-view?

Boyok1337 (approved):
Pay attention to urls naming

PR #872, opened 2025-04-17
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 
Based on requirements to the task. CinemaHall should use GenericViewSet. And do not use routes as for practice and understanding that you can do this too

PR #876, opened 2025-04-22
Completed task py-movie-api

Yurkiv313:


cinema/serializers.py
```diff
@@ -1,16 +1,78 @@
 from rest_framework import serializers
 
-from cinema.models import Movie
+from cinema.models import Movie, Actor, Genre, CinemaHall
+
+
+class GenreSerializer(serializers.Serializer):
```
pewlay, line 6: There were no restrictions in the task, so we could have used ModelSerializer, which would have reduced the code writing

cinema/views.py
```diff
@@ -1,45 +1,107 @@
...
 
-    if request.method == "DELETE":
-        movie.delete()
+    def delete(self, request, pk) -> Response:
+        self.get_object(pk=pk).delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class ActorList(
+    generics.ListAPIView,
```
pewlay, line 57: You could have used ListCreateAPIView, which would have reduced the code writing

pewlay (approved):
GJ! 🤩

Y-Havryliv (approved):

PR #878, opened 2025-04-23
add Actor, Genre, CinemaHall models. add actors and genres to Movie m…

yarzhuravlov:
…odel. add CRUD API routes and views for all models


BronkstoneBro (approved):
You're as great as always.

Y-Havryliv (approved):

PR #901, opened 2025-05-04
solution

hlibswork:


hlibswork:
@mateacademy-ai-mentor capacity field is not in the checklist

PR #975, opened 2025-08-30
'Solution'

DanilHamster:


cinema/models.py
```diff
@@ -5,6 +5,32 @@ class Movie(models.Model):
...
+
+    def __str__(self):
+        return f"Actor{self.first_name} {self.last_name}"
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=100, unique=True)
+
+    def __str__(self):
+        return self.name
```
MNDonut, line 27: you can move these models above to not use string in Movie model

MNDonut (approved):

PR #985, opened 2025-09-02
Solution

reinvvay:


cinema/urls.py
```diff
@@ -1,10 +1,50 @@
...
+movie_list = MovieViewSet.as_view(actions={
+    "get": "list",
+    "post": "create"
+})
+movie_detail = MovieViewSet.as_view(actions={
+    "get": "retrieve",
+    "put": "update",
+    "patch": "partial_update",
+    "delete": "destroy"
+})
```
Vskesha, line 34: You can use here DefaultRouter and register both CinemaHallViewSet and MovieViewSet. But this variant also works.

Vskesha (approved):
Clear solution. Everything is easy to understand. Code works properly and all required endpoints are implemented.

PR #998, opened 2025-09-02
Solution

viannik:


Hetmarum (approved):

viannik:
@mateacademy-ai-mentor In README it is written "Do not use related serializers for ManyToMany relations." So i haven't added them

PR #1074, opened 2025-12-27
Solution

stanislav-kishechko:


cinema/views.py
```diff
@@ -1,45 +1,133 @@
...
 
-@api_view(["GET", "PUT", "DELETE"])
-def movie_detail(request, pk):
-    movie = get_object_or_404(Movie, pk=pk)
+class GenreDetail(APIView):
+    def get_object(self, pk):
+        try:
+            return Genre.objects.get(pk=pk)
+        except Genre.DoesNotExist:
+            return None
```
Vorongor, line 45: Better option here is method- get_object_or_404

Vorongor (approved):

Vorongor (approved):
You missed an opportunity to use get_object_or_404 method, it allows you to skip "is None checks". Overall, Good Job!

PR #1077, opened 2025-12-28
feat: models, serializers, views, urls 1.0

AlexMishKo:


cinema/serializers.py
VadymVynokur, line 1: i think it can be improved, by using serializers.ModelSerializer

VadymVynokur (approved):
Well done!

PR #1086, opened 2025-12-29
Api Views

n-korobko:


vadimchubok (approved):
Your tests are failing. You need to rework the views so Luke can approve them 
Main problems in tests:
ImportError: cannot import name 'ActorList' from 'cinema.views'
ImportError: cannot import name 'GenreList' from 'cinema.views'

PR #1093, opened 2025-12-29
Solution

VYGAKOOL:


oleksii-shcherbak-code (commented):
Your tests fail because key classes are missing or misnamed.
The test suite cannot import GenreSerializer, ActorSerializer, CinemaHallSerializer, MovieSerializer, or MovieViewSet.
Add these classes to cinema/serializers.py and cinema/views.py with the exact names expected by the tests.

oleksii-shcherbak-code (changes requested):
Your tests fail because key classes are missing or misnamed.

PR #1145, opened 2026-04-08
Solution

idubina:


idubina:
@mateacademy-ai-mentor
There is a task condition in read me file: "Create serializers for all these models. Do not use related serializers for ManyToMany relations." That is why I didn't add ManyToManyField relationships to the serializer

PR #1261, opened 2026-08-14
Complete

Rostyk-D:


cinema/views.py
```diff
@@ -1,45 +1,136 @@
...
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    generics.GenericAPIView,
+):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+    def get(self, request, pk):
+        return self.retrieve(request, pk)
```
Sn1kls: Optional style note: `pk` is passed positionally here, but the mixin methods (`retrieve`/`update`/`partial_update`/`destroy`) read the lookup from `self.kwargs`, so this `pk` argument is actually ignored. The idiomatic form is to forward `*args`/`**kwargs`:

```
def get(self, request, *args, **kwargs):
    return self.retrieve(request, *args, **kwargs)
```

Applies to `put`/`patch`/`delete` here too. Works fine as-is – just cleaner.

Sn1kls (approved):
Approved.

Sn1kls (approved):

